PR #1, opened 2022-04-22
DM-34450: Move RubinTV to its own package

mfisherlevine:


ups/rubintv_production.cfg
```diff
@@ -0,0 +1,20 @@
...
+}
+
+# For packages that build a C++ library and a Python module, the below should be sufficient.
+# Pure-Python packages should set headers=[], libs=[] (not libs=None). and hasSwigFiles=False.
+# For more information, see the sconsUtils Doxygen documentation.
+config = lsst.sconsUtils.Configuration(
+    __file__,
+    headers=[],
+    hasDoxygenInclude=False,
+)
```
TallJimbo, line 20: This file should be unnecessary for a pure-Python package.

TallJimbo (approved):

ups/rubintv_production.cfg
mfisherlevine: As discussed OOB, this does seem to be necessary 😕

PR #2, opened 2022-04-22
DM-34315: Move rapid_analysis code to summit_utils and summit_extras

mfisherlevine:


ups/rubintv_production.table
```diff
@@ -5,8 +5,9 @@
 setupRequired(atmospec)
 setupRequired(base)
 setupRequired(utils)
-setupRequired(rapid_analysis)
 setupRequired(pex_exceptions)
+setupRequired(summit_utils)
+setupRequired(summit_extras)
```
isullivan, line 10: Is `summit_extras` actually required?

isullivan (approved):
You need to remove the "DO NOT MERGE ME" commit...

ups/rubintv_production.table
mfisherlevine: Hmmm, right now, no, but I thought it wasn't unlikely. But that's fair, I'll remove it for now, can always add it later (I just have forgotten to do so in the past, but that's not a good reason to deliberately get it wrong now!)

PR #3, opened 2022-04-26
DM-34238: Make use of new butler init API

mfisherlevine:


python/lsst/rubintv/production/rubinTv.py
```diff
@@ -402,14 +401,14 @@ class MountTorqueChannel():
     """Class for running the mount torque channel on RubinTV.
     """
 
-    def __init__(self, location, doRaise=False):
+    def __init__(self, doRaise=False):
```
czwa, line 404: I think I've previously complained that these could derive from a base class to handle all the duplicate code, so I'd like to be consistent and say the same thing again.

czwa (approved):

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: Haha fair enough. You were probably right then, and you're probably right now, but this PR isn't about that 🙂

PR #4, opened 2022-04-27
DM-34582: Add util for setting up logging and apply to RubinTV scripts

mfisherlevine:


python/lsst/rubintv/production/utils.py
```diff
@@ -72,3 +73,8 @@ def checkRubinTvExternalPackages(exitIfNotFound=True, logger=None):
 
     if exitIfNotFound and (not hasGoogleStorage or not hasEfdClient):
         exit()
+
+
+def setupLogging(longlog=False):
```
parejkoj, line 78: Please add a docstring.

parejkoj (commented):
My initial reaction was to say that we should have this method available somewhere for the rest of the stack, but it's only two lines so we probably don't need that.

Give it a docstring and I think it's good.

python/lsst/rubintv/production/utils.py
mfisherlevine: Move to summit_utils and added.

PR #8, opened 2022-05-17
DM-34631: Add cleanup for GCS bucket

mfisherlevine:


czwa (approved):
The two blocks seem like they do nearly the same thing, and could probably be refactored to reduce code, but that can be a future ticket.

mfisherlevine:
Thanks for the speedy review. Just to note that are indeed quite similar, but some also somewhat different, as one is excluding all the final movies, and I felt like they were more likely to diverge than converge, so rather than trying to get fancy and accidentally deleting things I didn't want to, just thought I'd copy & paste a bit.

PR #10, opened 2022-06-15
DM-35208: Increase polynomial order in mount tracking plots

craiglagegit:
This is ready to merge.  This comment is in the ticket:

It turns out that the problem was not just the higher order polynomial, but that the fits were poorly conditioned because the time variable was a large number. Subtracting an offset so that the time variable is centered in the window fixed this. Pushing the fix now. See the attached comparisons on a 5 minute exposure from 20220526.

Also attaching the before and after plots.
[Mount_Tracking_2022052600012.pdf](https://github.com/lsst-sitcom/rubintv_production/files/8911737/Mount_Tracking_2022052600012.pdf)
[Mount_Tracking_Old_2022052600012.pdf](https://github.com/lsst-sitcom/rubintv_production/files/8911738/Mount_Tracking_Old_2022052600012.pdf)

### DIFF
--- a/python/lsst/rubintv/production/mountTorques.py
+++ b/python/lsst/rubintv/production/mountTorques.py
@@ -33,7 +33,10 @@ try:
 except ImportError:
     pass
 
-GOOD_IMAGE_TYPES = ['OBJECT', 'SKYEXP', 'ENGTEST', 'SCIENCE', 'ACQ', 'CWFS']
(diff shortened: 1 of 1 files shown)


mfisherlevine:
Thanks for the help here Craig, that's great. I've reviewed, and we spoke OOB to check the plots etc. I've made the necessary code corrections myself (was just flake8), and added a couple little tweaks to some logic too. Merging now.

PR #14, opened 2022-07-01
DM-35398: Add seeing and focus info to table data

mfisherlevine:


czwa (approved):
Are these keywords guaranteed to exist in the raw metadata?

mfisherlevine:
> Are these keywords guaranteed to exist in the raw metadata?

These days, yes, I think so, but honestly I did have the same thought, plus I'd like it to just _always_ work, so I'll redo that with a check.

PR #15, opened 2022-07-07
DM-35445: Write background catchup service

mfisherlevine:


python/lsst/rubintv/production/endOfDay.py
```diff
@@ -0,0 +1,281 @@
...
+                                                                where=where,
+                                                                bind={'dayObs': self.dayObs},
+                                                                datasets='quickLookExp')
+        expRecords = list(set(expRecords))
+        foundSeqNums = [r.seq_num for r in expRecords]
+        toMakeSeqNums = [s for s in allSeqNums if s not in foundSeqNums]
+        return [{'day_obs': self.dayObs, 'seq_num': s, 'detector': 0} for s in toMakeSeqNums]
+
+    @staticmethod
+    def _minimizeDataId(dataId):
```
czwa, line 146: This name is confusing to me.  I think part of the reason is that this is doing the opposite of what `daf_butler`'s "expanded" dataIds do (although both "shrink" and "compress" seem like bad verbs too).

python/lsst/rubintv/production/endOfDay.py
```diff
@@ -0,0 +1,281 @@
...
+        finally:
+            self.dayObs = getCurrentDayObs_int()
+
+    def run(self):
+        """Runs forever, running the catchup services during the day
+        and the end of day service when the day ends.
+
+        Raises
+        ------
+        RuntimerError:
```
czwa, line 260: Typo.

python/lsst/rubintv/production/utils.py
```diff
@@ -210,12 +213,21 @@ def remakeDay(channel, dayObs, remakeExisting=False, notebook=True, **kwargs):
     ------
     ValueError:
         Raised if the channel is unknown.
+        Raised if remakeExisting is False and channel is auxtel_metadata
```
czwa, line 216: Ending period.

python/lsst/rubintv/production/utils.py
```diff
@@ -210,12 +213,21 @@ def remakeDay(channel, dayObs, remakeExisting=False, notebook=True, **kwargs):
...
     """
+    if not logger:
+        logger = logging.getLogger(__name__)
+
     from google.cloud import storage
 
     if channel not in CHANNELS:
         raise ValueError(f"Channel {channel} not in {CHANNELS}")
 
+    if remakeExisting and channel in ['auxtel_metadata']:
```
czwa, line 226: Am I thinking about this logic wrong?  If `remakeExisting = True`, then this block will trigger, but it claims that `remakeExisting = True` will avoid this?

scripts/runBackgroundService.py
```diff
@@ -0,0 +1,38 @@
...
+
+allSkyPngRoot = '/project/rubintv/allsky/'
+moviePngRoot = '/project/rubintv/auxTel_animation/'
+
+setupLogging()
+checkRubinTvExternalPackages()
+
+print('Running RubinTV background catchup service...')
+# TODO: change dryRunForDeletion for deployment, but test like this first
+allSky = RubinTvBackgroundService(allSkyPngRoot=allSkyPngRoot,
```
czwa, line 34: Should these be `argparse` options, so the script doesn't need to be edited to switch modes?

czwa (approved):
Mostly docstring/typo issues, with one potential logic error.

scripts/runBackgroundService.py
mfisherlevine: These are actually entry points for the k8s pods which run these, so in this case it would mean adding complexity only to have to recode the k8s runners to put in the args correctly (plus all the channels do it like this).

Really, I will probably just remove `dryRun` entirely once I've seen it print out the correct path - I just put it in because launching a new service and letting it run `shutil.rmtree()` on some path without being _really_ sure it's what you want is not a good plan 🙃

python/lsst/rubintv/production/utils.py
mfisherlevine: Nope, you are 100% correct, that logic is backwards and the docs are correct. Fixed to `if remakeExisting is False...`

python/lsst/rubintv/production/endOfDay.py
mfisherlevine: Wow, good eyes!

python/lsst/rubintv/production/endOfDay.py
mfisherlevine: Yup, agreed, it's awful. I've changed it to `_makeMinimalDataId()` because I think that's most explicit. Let me know if that's still distasteful though.

python/lsst/rubintv/production/endOfDay.py
```diff
@@ -0,0 +1,281 @@
...
+        if currentDay == self.dayObs:
+            return False
+        elif currentDay == self.dayObs+1:
+            return True
+        else:
+            raise RuntimeError(f"Encountered non-linear time! Day cleaner was started with {self.dayObs}"
+                               f"and now current dayObs is {currentDay}!")
+
+    def getMissingQuickLookIds(self):
+        """XXX DOCS
```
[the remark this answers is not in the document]
mfisherlevine: Yeah, sorry about that. All done now though.

python/lsst/rubintv/production/endOfDay.py
```diff
@@ -0,0 +1,281 @@
...
+
+    Parameters
+    ----------
+    allSkyPngRoot : `str`
+        The path at which the all sky movie channel is writing its images to.
+    moviePngRoot : `str`
+        The root path to write all pngs and movies to. It need not exist,
+        but must be creatable with write privileges.
+    doRaise `bool`
+        Raise on error?
```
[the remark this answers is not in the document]
mfisherlevine: As discussed OOB, removed now that it's seen to be safe known to be safe.

PR #16, opened 2022-07-07
DM-35486: Add image impact of errors

craiglagegit:
This change adds the image impact of the tracking errors.  The AZ error is scaled by cos(EL) and the rotator error is calculated as r*theta, where r is the angular distance to the corner of the field (280 arcseconds for AuxTel) and theta is the rotator tracking error.  Attached is a graph with the change:
[Mount_Torques_2022052600012.pdf](https://github.com/lsst-sitcom/rubintv_production/files/9065275/Mount_Torques_2022052600012.pdf)

### DIFF
--- a/python/lsst/rubintv/production/mountTorques.py
+++ b/python/lsst/rubintv/production/mountTorques.py
@@ -192,7 +192,7 @@ def calculateMountErrors(dataId, butler, client, figure, saveFilename, logger):
         plt.subplot(3, 3, 4)
         plt.plot(fit_times, az_error, color='red')
 
-        plt.title(f"Azimuth RMS error = {az_rms:.2f} arcseconds\n" + \
+        plt.title(f"Azimuth RMS error = {az_rms:.2f} arcseconds\n"
                   f"  Image RMS error = {image_az_rms:.2f} arcseconds")
         plt.ylim(-10.0, 10.0)
         plt.xticks([])
@@ -211,7 +211,7 @@ def calculateMountErrors(dataId, butler, client, figure, saveFilename, logger):
         ax2.set_xticks([])
         plt.subplot(3, 3, 5)
         plt.plot(fit_times, el_error, color='green')
-        plt.title(f"Elevation RMS error = {el_rms:.2f} arcseconds\n" + \
+        plt.title(f"Elevation RMS error = {el_rms:.2f} arcseconds\n"
                   f"    Image RMS error = {image_el_rms:.2f} arcseconds")
         plt.ylim(-10.0, 10.0)
         plt.xticks([])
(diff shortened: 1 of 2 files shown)


python/lsst/rubintv/production/mountTorques.py
```diff
@@ -164,68 +168,77 @@ def plotMountTracking(dataId, butler, client, figure, saveFilename, logger):
...
+        figure.clear()
+        title = f"Mount Tracking {dataIdString}, Azimuth = {azimuth:.1f}, Elevation = {elevation:.1f}"
+        plt.suptitle(title, fontsize=18)
+        # Azimuth axis
+        plt.subplot(3, 3, 1)
+        ax1 = az['azimuthCalculatedAngle'].plot(legend=True, color='red')
+        ax1.set_title("Azimuth axis", fontsize=16)
+        ax1.axvline(az.index[0], color="red", linestyle="--")
+        ax1.set_xticks([])
+        ax1.set_ylabel("Degrees")
```
czwa, line 191: Can the x-axes be labeled/given units?  It looks like they're all times, except for the Nasmyth RMS error (which is degrees).

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -455,7 +455,7 @@ def callback(self, dataId):
         try:
             tempFilename = tempfile.mktemp(suffix='.png')
             uploadFilename = _dataIdToFilename(self.channel, dataId)
-            plotted = plotMountTracking(dataId, self.butler, self.client, self.fig, tempFilename, self.log)
+            plotted = calculateMountErrors(dataId, self.butler, self.client, self.fig, tempFilename, self.log)
```
czwa, line 458: `calculateMountError` returns the list of arrays discussed above, which will always be True-like and run the following block.  Is that intended?

czwa (approved):
Minor clarity suggestions/questions of things that are just not obvious to me.

python/lsst/rubintv/production/mountTorques.py
```diff
@@ -164,68 +168,77 @@ def plotMountTracking(dataId, butler, client, figure, saveFilename, logger):
     el_rms = np.sqrt(np.mean(el_error * el_error))
     rot_rms = np.sqrt(np.mean(rot_error * rot_error))
 
+    # Calculate Image impact RMS
+    image_az_rms = az_rms * np.cos(el_vals[0] * np.pi / 180.0)
+    image_el_rms = el_rms
+    image_rot_rms = rot_rms * 280.0 * np.pi / 180.0 / 3600.0
```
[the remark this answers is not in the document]
mfisherlevine: As discussed OOB, hard-coded as an obviously AuxTel specific const for now `AUXTEL_ANGLE_TO_EDGE_OF_FIELD_ARCSEC`

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: Nice catch, thanks. Have changed to just test if the output file was written.

python/lsst/rubintv/production/mountTorques.py
mfisherlevine: They are actually all sharing the bottom axis, and they are indeed all time, which I think is obvious enough at the bottom. However, for some reason I can't quite work out, this plot is showing its axis (which is actually seconds, centered on zero), an so looks different (the bottom is absolute time, the intermediate plots are relative time, centered on the observation midpoint for fitting reasons).

I just don't know why this one is showing its ticks when the others aren't and they're all turned off the same way...

mfisherlevine: Ah, caught it. Removing that random addition of ticks and I think we're done.

PR #17, opened 2022-07-08
DM-35473: Add heartbeats to RubinTV

mfisherlevine:


python/lsst/rubintv/production/rubinTv.py
```diff
@@ -141,6 +144,30 @@ def __init__(self):
...
+        heartbeatJson = json.dumps(heartbeatJsonDict)
+
+        blob = self.bucket.blob("/".join([filename]))
+        blob.cache_control = 'no-store'  # must set before upload
+        blob.upload_from_string(heartbeatJson)
+
+        self.log.debug(f'Uploaded heartbeat to channel {channel} with datetime {currTime}')
+
+        blob.reload()
+        if blob.cache_control != 'no-store':
```
czwa, line 168: I realize this isn't supposed to trigger, but is it possible to trigger?  Does `blob.reload()` potentially alter the cache_control property?

czwa (approved):
I think a lot of the code duplication could be removed by making the heartbeat into a separate class, but this looks reasonable otherwise.

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: It certainly shouldn't be able to trigger, no. Given that I've never seen it do so (it was mainly put in as an assert type statement during development while we were trying to get the caches to behave) it could probably safely come out, but we also don't care about speed here at all, and if it were ever to fire I'd quite like the error to be raised, so I think leaving it is probably best on balance.

mfisherlevine: Plus in the not so distant future we'll be moving from GCS to S3, so all of this will have to go at that point.

PR #18, opened 2022-07-12
DM-35545: Add util to test current channels when telescope is offline

mfisherlevine:


python/lsst/rubintv/production/utils.py
```diff
@@ -258,3 +262,63 @@ def remakeDay(channel, dayObs, remakeExisting=False, notebook=True, logger=None,
...
+    for ``duration`` and then deletes the file.
+
+    Parameters
+    ----------
+    channel : `str`
+        The name of the lsst.rubintv.production channel. The actual channel
+        object is created internally.
+    duration : `float`, optional
+        The duration to leave the test image up for, in seconds.
+    """
```
czwa, line 285: This should have a `Raises` section for the `ValueError`s.

python/lsst/rubintv/production/utils.py
```diff
@@ -258,3 +262,63 @@ def remakeDay(channel, dayObs, remakeExisting=False, notebook=True, logger=None,
...
+    logger.info(f'Found {len(blobs)} for channel {channel} in bucket')
+
+    # names are like
+    # 'auxtel_monitor/auxtel-monitor_dayObs_2021-07-06_seqNum_100.png'
+    days = set([b.name.split(f'{prefix}_dayObs_')[1].split('_seqNum')[0] for b in blobs])
+    days = [int(d.replace('-', '')) for d in days]  # days are like 2022-01-02
+    recentDay = max(days)
+
+    seqNums = getPlotSeqNumsForDayObs(channel, recentDay, bucket)
+    newSeqNum = max(seqNums) + 1
```
czwa, line 311: I'm a bit worried that `+ 1` may allow a race condition between this test image and potential new incoming images.  If everything blocks, so no update can happen, then this is fine.  If this is something that may prevent an image being generated with that `newSeqNum` value, and image generation will resume after this test file is removed (the catchup code for example), then it's also fine.

czwa (approved):
I'm a bit concerned that there may be a race condition between this code and the background image generation.  I think I've convinced myself that the catch-up code makes this self-healing, but I just wanted to prompt one more check on your side to make sure this isn't an issue.

python/lsst/rubintv/production/utils.py
mfisherlevine: Oh no, you're 100% right about the race conditions, and overwriting images etc. This code is absolutely *only* for use when things aren't online and taking data! When things are rolling in then we have images flowing to test the frontend code, and wouldn't need to use this. This is entirely for developer use to test when we don't have any real data flowing.

mfisherlevine: Given the utility functions that now exist for regenerating whole days etc, plus the catchup services, and the intention that this is only used by experts, I personally think that this is all fine though. If you disagree please let me know though.

PR #20, opened 2022-09-14
DM-36204: Add alt to metadata table

mfisherlevine:


python/lsst/rubintv/production/rubinTv.py
```diff
@@ -530,10 +530,13 @@ def dataIdToMetadataDict(butler, dataId, keysToRemove):
         rawmd = butler.get('raw.metadata', dataId)
         obsInfo = ObservationInfo(rawmd)
         d['airmass'] = obsInfo.boresight_airmass
+        d['focus_z'] = obsInfo.focus_z.value
 
-        if 'FOCUSZ' in rawmd:
-            d['focus_z'] = rawmd['FOCUSZ']
-        if 'SEEING' in rawmd:
+        d['Altitude'] = 'null'  # altaz_begin is None when not on sky so need check it's not None first
```
timj, line 535: Assigning a string to something that should be a float does make me wince a bit when thinking about a `mypy` future. Is there a reason this is not `None` or `NaN`?

timj (approved):
Looks okay to me. I don't understand the `"null"` usage though.

timj:
You really need to add some branch protections and GitHub actions to this repo, even if it's just copying over the lint.yaml from science pipelines.

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: Yeah, fair enough. I think `None` is the right answer here, I just need to check that that does indeed get rendered as `null` like everything else on the front end. I'll do that tomorrow morning once we're not on sky, and will merge with that change assuming it works. I don't think any of this will ever be mypyed though.

mfisherlevine:
> You really need to add some branch protections and GitHub actions to this repo, even if it's just copying over the lint.yaml from science pipelines.

Yeah, this is a very good shout. I'll take a look at that in the morning as well.

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: `None` works perfectly. Merging now.

PR #25, opened 2022-10-06
DM-36363: Add retries to GCS uploads

mfisherlevine:


python/lsst/rubintv/production/rubinTv.py
```diff
@@ -207,10 +226,14 @@ def googleUpload(self, channel, sourceFilename, uploadAsFilename=None, isLiveFil
...
-        if isLiveFile:
-            blob.reload()
-            assert(blob.cache_control == 'no-store')
+
+        # general retry strategy
+        # still quite gentle as the catchup service will fill in gaps
+        # and we don't want to hold up new images landing
+        modified_retry = DEFAULT_RETRY.with_deadline(2.0)  # in seconds
+        modified_retry = modified_retry.with_delay(initial=.5, multiplier=1.2, maximum=2)
+        blob.upload_from_filename(finalName, retry=modified_retry)
```
czwa, line 235: Do you care if this raises?  I see the prior upload in `uploadHeartbeat` has the try/except protection, and wanted to confirm it doesn't need such here.

czwa (approved):

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: Yeah, I think that's a good point. It doesn't _need_ one, per se, but putting one in will avoid unnecessary container restarts (though they should become much more rare anyway, as the retries should reduce the number of random failures quite a bit, hopefully).

PR #26, opened 2022-10-21
DM-36563: Refactor heartbeat code into its own class

mfisherlevine:
Add ability to do per-call forecasting too.


python/lsst/rubintv/production/heartbeat.py
```diff
@@ -0,0 +1,76 @@
...
+
+        customFlatlinePeriod implies ensure, as long forecasts should always be
+        uploaded.
+
+        Parameters
+        ----------
+        ensure : `str`
+            Ensure that the heartbeat is uploaded, even if one has been sent
+            within the last ``uploadPeriod``.
+        customFlatlinePeriod : `float`
```
czwa, line 66: I'm not sure I understand this parameter.  Above, you use it in the catchup script, but it's `flatlinePeriod`, not `customFlatlinePeriod`.  I'm also not seeing where the flatline period is being used.  Is there a function I'm missing that checks `lastUpload` against `flatlinePeriod` (either the class or custom one), and signals that there's an error?

mfisherlevine: Ah, good catch - that's a typo, that should indeed be `customFlatlinePeriod` in those calls, thanks. Locally it's being used to set the forecast, _i.e._ the time after which the channel is considered down.

On the website, the heartbeats' timestamps are being checked that they're always coming before this "considered-dead time", so that you can keep sending them and pushing that time out, but can also have long-running things, where you say "You won't hear from me for an hour, but that's OK in this instance because I'm doing something very long & slow".

czwa: Ok, the polling code is elsewhere, that's why I didn't see it.  Makes sense.  I have no other issues, so will mark this as approved.

czwa (approved):

PR #27, opened 2022-10-31
DM-36727: Change metadata handling to allow partial uploads

mfisherlevine:
This also means we can add supplementary info for each seqNum post hoc.


python/lsst/rubintv/production/rubinTv.py
```diff
@@ -554,3 +631,161 @@ def run(self):
...
+            The channel name.
+        flatlinePeriod : `float`
+            The period after which to consider the channel dead.
+
+        Returns
+        -------
+        success : `bool`
+            Did the upload succeed?
+        """
+        filename = "/".join([self.HEARTBEAT_PREFIX, channel]) + ".json"
```
czwa, line 664: Are these real paths?  Would `os.path.join` be better here and below?

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -554,3 +631,161 @@ def run(self):
...
+
+        # general retry strategy
+        # still quite gentle as the catchup service will fill in gaps
+        # and we don't want to hold up new images landing
+        modified_retry = DEFAULT_RETRY.with_deadline(2.0)  # in seconds
+        modified_retry = modified_retry.with_delay(initial=.5, multiplier=1.2, maximum=2)
+        try:
+            blob.upload_from_filename(finalName, retry=modified_retry)
+        except Exception:
+            return None
```
czwa, line 739: I feel like I may have asked this previously on another ticket: do you not care about the exception because this will try again later?  Just wondering if a log message for either case is helpful.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -554,3 +631,161 @@ def run(self):
...
+
+        return blob
+
+
+class Heartbeater():
```
czwa, line 744: I'm almost positive I've suggested a base class for these, and am now suggesting it again for a future update.

czwa (approved):
A few comments for clarification, and a possible typo that may be confusing in the future.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -48,9 +58,8 @@
 
 from lsst.rubintv.production.mountTorques import calculateMountErrors
 from lsst.rubintv.production.monitorPlotting import plotExp
-from .uploadTools import Heartbeater, Uploader
-from .channels import PREFIXES
-
+from .channels import PREFIXES, CHANNELS
+from .utils import writeMetataShard, isFileWorldWritable
```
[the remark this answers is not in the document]
mfisherlevine: Oh good lord, gross, thank you!

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -62,6 +71,31 @@
...
+                "sky_angle": 'Sky angle',
+                "azimuth": 'Azimuth',
+                "zenith_angle": 'Zenith angle',
+                "time_begin_tai": 'TAI',
+                "filter": 'Filter',
+                "disperser": 'Disperser',
+                "airmass": 'Airmass',
+                "focus_z": 'Focus-Z',
+                "seeing": 'DIMM Seeing',
+                "altitude": 'DIMM Seeing',
```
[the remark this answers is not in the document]
mfisherlevine: Yup, thanks!

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: I'm actually conflicted. They're not "real" paths in the strict/traditional sense, as they're the names of `blob`s in GCS, but these do basically function just like paths, though I forget the rules on leading slashes. I think I'd prefer not to change this right now, just in case, but am happy to have a big tidy up of things once we have a TTS version of RubinTV running, if that's OK.

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: Correct, the catchup service will try again later, but yes, I agree, no harm in logging a warning, I'll add one now.

python/lsst/rubintv/production/catchupService.py
```diff
@@ -254,10 +254,18 @@ def runCatchup(self):
         """
         startTime = time.time()
 
-        self.catchupMetadata()
-        self.catchupIsrRunner()
-        self.catchupMonitor()
-        self.catchupMountTorques()
+        # a little ugly but saves copy/pasting the try block 4 times
+        # we need to try each one because raising here has bad conqquences
```
czwa, line 258: Typo on consequences.

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: I thought you just suggested using _a class at all_, and now it is one. It used to be functions copied and pasted around, and I was also offended by that, and now it's a class that each object gets, and there's no gross `nonlocal` nonsense anywhere.

czwa (approved):

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: Or did you not mean the `Heartbeater`, and that every channel starts with a base class? I'm honestly still on the fence about that, given how different they are, and how they might continue to grow apart, but I certainly take your point, I feel that pull too, I'm just still not certain it's net-positive.

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: Probably somewhere in between, tbh 🙃

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -480,69 +520,106 @@ def dataIdToMetadataDict(butler, dataId, keysToRemove):
...
+            data = {}
+            # json.load() doesn't like empty files so check size is non-zero
+            if os.path.isfile(mainFile) and os.path.getsize(mainFile) > 0:
+                with open(mainFile) as f:
+                    data = json.load(f)
+
+            with open(shardFile) as f:
+                shard = json.load(f)
+            if shard:
+                for row in shard:
```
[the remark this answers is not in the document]
mfisherlevine: Almost. So, `data` is the contents of the main file, which the shard's contents are being merged into - that's just been loaded from disk from `mainFile`. Each shard can be one or more rows (where a row is a seqNum), and each row is a dict of metadata with one or more entries. So the minimal shard is a single k,v pair for a single seqNum, but it could also be 20 k,v pairs for 150 seqNums, for example, and there could also be multiple shards, all being merged into the main file.

You are quite right about the potential for this automatically taking the last entry and dropping all others, but this is for filling gaps in tables and adding new entries. If we're re-writing items here something has gone desperately wrong with the other services and this is the least of our worries.

python/lsst/rubintv/production/rubinTv.py
czwa: I think I confused myself on what the Heartbeat code looked like in previous tickets.

PR #29, opened 2022-11-14
DM-36972: Add mount image degradation metrics to LATISS page

mfisherlevine:


python/lsst/rubintv/production/utils.py
```diff
@@ -282,7 +286,7 @@ def remakeDay(channel, dayObs, *,
 
     # doRaise is False because during bulk plot remaking we expect many fails
     # due to image types, short exposures, etc.
-    tvChannel = createChannelByName(channel, doRaise=True, embargo=embargo, **kwargs)
+    tvChannel = createChannelByName(channel, doRaise=False, embargo=embargo, **kwargs)
```
mfisherlevine, line 289: Reverting accidental change of this default in `de48877454fd`.

python/lsst/rubintv/production/mountTorques.py
```diff
@@ -39,6 +39,8 @@
                             ]
 
 AUXTEL_ANGLE_TO_EDGE_OF_FIELD_ARCSEC = 280.0
+MOUNT_IMAGE_WARNING_LEVEL = .25  # this determines the colouring of the cells in the table, yellow for this
```
czwa, line 42: It might be good to have an __all__ to control what gets exported (as it's * in __init__.py).

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -422,10 +423,32 @@ def __init__(self, doRaise=False):
...
+    def writeMountErrorShard(self, errors, dataId):
+        """Write a metadata shard for the mount error, including the flag
+        for coloring the cell based on the threshold values.
+        """
+        dayObs = butlerUtils.getDayObs(dataId)
+        seqNum = butlerUtils.getSeqNum(dataId)
+        az_rms, el_rms, _ = errors  # we don't need the rot errors here so assign to _
+        imageError = (az_rms ** 2 + el_rms ** 2) ** .5
+        key = 'Mount motion image degradation'
+        flagKey = '_' + key  # color coding of cells always done by prepending with an underscore
```
czwa, line 440: This key is only added if there's an issue.  Should this be always populated, with a key like "Mount motion quality" `in  {'good', 'warning', 'bad'}`?

czwa (approved):
Two small questions, neither of which should be blocking.

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: I think, for now, we don't want to add the good by default. The front end will colour that green, and I think, for now, we just want to leave it grey, but it's certainly potentially to be added in the future, thanks.

PR #32, opened 2022-12-06
DM-37252: Move util from RubinTV to summit_utils

mfisherlevine:


isullivan (approved):
Code removal looks fine

PR #33, opened 2022-12-14
DM-37091: Add exposure summary stats to RubinTV

mfisherlevine:


python/lsst/rubintv/production/rubinTv.py
```diff
@@ -825,3 +850,201 @@ def beat(self, ensure=False, customFlatlinePeriod=None):
...
+class CalibrateCcdRunner():
+    """Class for running CharacterizeImageTask and CalibrateTasks on images.
+
+    Runs these tasks and writes shards with various measured quantities for
+    upload to the table.
+    """
+    def __init__(self, outputRoot, *, doRaise=False, embargo=False):
+        self.dataProduct = 'quickLookExp'
+        self.uploader = Uploader()
+        self.butler = makeDefaultLatissButler(embargo=embargo, writeable=True)
```
mfisherlevine, line 864: Add a comment to say that `writeable=True` is required to define the visits.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -825,3 +850,201 @@ def beat(self, ensure=False, customFlatlinePeriod=None):
...
+        self.log = _LOG.getChild("monitorChannel")
```
mfisherlevine, line 865: Logger needs renaming

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -825,3 +850,201 @@ def beat(self, ensure=False, customFlatlinePeriod=None):
...
+        self.channel = 'auxtel_characterizeImage'
```
mfisherlevine, line 866: 1) This should probably be renamed to match the new class name.
2) If this ever actually uploads a plot itself (and it may well not) the this string will need adding to the list in `channels.py`.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -825,3 +850,201 @@ def beat(self, ensure=False, customFlatlinePeriod=None):
...
+        self.watcher = Watcher(self.dataProduct, self.channel)
+        self.fig = plt.figure(figsize=(12, 12))
```
mfisherlevine, line 868: Currently unused.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -825,3 +850,201 @@ def beat(self, ensure=False, customFlatlinePeriod=None):
...
+        self.doRaise = doRaise
+        self.shardsDir = os.path.join(outputRoot, 'shards')
+        self.outputRunName = "LATISS/runs/quickLook/1"
```
mfisherlevine, line 871: This needs taking from somewhere more central. DM-37272 exists for this, so please add a `TODO: DM-37272` to remind me/help me find it.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -825,3 +850,201 @@ def beat(self, ensure=False, customFlatlinePeriod=None):
...
+        config = CharacterizeImageConfig()
+        basicConfig = CharacterizeImageConfig()
+        obs_lsst = getPackageDir("obs_lsst")
+        config.load(os.path.join(obs_lsst, "config", "characterizeImage.py"))
+        config.load(os.path.join(obs_lsst, "config", "latiss", "characterizeImage.py"))
+        config.measurement = basicConfig.measurement
+
+        config.doApCorr = False
+        config.doDeblend = False
+        self.charImage = CharacterizeImageTask(config=config)
```
mfisherlevine, line 882: Let's file a ticket for testing out some further performance tweaks, such as going down to a single round of PSF estimation, reducing the number of sources measured in busy fields, other config options to turn off totally.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -825,3 +850,201 @@ def beat(self, ensure=False, customFlatlinePeriod=None):
...
+        config.measurement = basicConfig.measurement
+
+        config.doApCorr = False
+        config.doDeblend = False
+
+        self.calibrate = CalibrateTask(config=config, icSourceSchema=self.charImage.schema)
+
+    def _getRefObjLoader(self, refcatName, dataId, config):
+        """Get a reference object loader for a given refcat and dataId.
+        """
```
mfisherlevine, line 897: Could do with a better docstring. And I think that's like the case in enough places that I'll just say it once here.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -825,3 +850,201 @@ def beat(self, ensure=False, customFlatlinePeriod=None):
...
+        refs = self.butler.registry.queryDatasets(refcatName, dataId=dataId).expanded()
+        refs = list(refs)
```
mfisherlevine, line 899: Given that this wrapping in `list()` is _critical_ for this to work consistently, I think it needs a comment to say so, lest generator-lover "optimize" this out in future.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -825,3 +850,201 @@ def beat(self, ensure=False, customFlatlinePeriod=None):
...
+
+            self.log.info(f'Running Image Characterization for {dataId}')
+            exp = _waitForDataProduct(self.butler, self.dataProduct, dataId, self.log)
+
+            if not exp:
+                raise RuntimeError(f'Failed to get {self.dataProduct} for {dataId}')
+
+            if isDispersedExp(exp):
+                self.log.info(f'Skipping dispersed image: {dataId}')
+                return
```
mfisherlevine, line 931: Let's also file a ticket for both turning this off to see how well it can manage dispersed images, and also stopping it flailing on calibs. These can all be filed on me, of course.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -825,3 +850,201 @@ def beat(self, ensure=False, customFlatlinePeriod=None):
...
+
+            charRes = self.charImage.run(exp)
+            tCharacterize = time.time()
+            self.log.info(f"Ran characterizeImageTask in {tCharacterize-tStart:.2f} seconds")
+
+            nSources = len(charRes.sourceCat)
+            dayObs = butlerUtils.getDayObs(dataId)
+            seqNum = butlerUtils.getSeqNum(dataId)
+            outputDict = {"50-sigma source count": nSources}
+            mdDict = {seqNum: outputDict}
```
mfisherlevine, line 951: Add the `_measured` flag here, as this is the odd one out in the table now.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -825,3 +850,201 @@ def beat(self, ensure=False, customFlatlinePeriod=None):
...
+            writeMetadataShard(self.shardsDir, dayObs, mdDict)
+
+            calibrateRes = self.calibrate.run(charRes.exposure,
+                                              background=charRes.background,
+                                              icSourceCat=charRes.sourceCat)
+            tCalibrate = time.time()
+            self.log.info(f"Ran calibrateTask in {tCalibrate-tCharacterize:.2f} seconds")
+
+            summaryStats = calibrateRes.outputExposure.getInfo().getSummaryStats()
+            pixToArcseconds = calibrateRes.outputExposure.getWcs().getPixelScale().asArcseconds()
```
mfisherlevine, line 961: I think we should grab the wcs above before the fit and report the change in the wcs here too, including the change in the rotation angle. Can be another ticket if you don't want to do it on this one, but it would be nice to add at this point I think.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -825,3 +850,201 @@ def beat(self, ensure=False, customFlatlinePeriod=None):
...
+                'Zeropoint': summaryStats.zeroPoint
+            }
+
+            # flag all these as measured items to color the cell
+            labels = {"_" + k: "measured" for k in outputDict.keys()}
+            outputDict.update(labels)
+
+            mdDict = {seqNum: outputDict}
+            writeMetadataShard(self.shardsDir, dayObs, mdDict)
+            self.log.info(f'Wrote metadata shard. Putting calexp. {dataId}')
```
mfisherlevine, line 986: `"Wrote metadata shard. Putting calexp for {dataId}"`

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -825,3 +850,201 @@ def beat(self, ensure=False, customFlatlinePeriod=None):
...
+            self.safeWrite(calibrateRes.outputExposure, "calexp", dataId)
+            tFinal = time.time()
+            self.log.info(f"Ran everything in {tFinal-tStart:.2f} seconds")
```
mfisherlevine, line 989: Maybe change "everything" to either characterize + calibrate, or "single frame processing" or something?

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -825,3 +850,201 @@ def beat(self, ensure=False, customFlatlinePeriod=None):
...
+        visitDataIds = list(set(visitDataIds))
+        if len(visitDataIds) == 1:
+            visitDataId = visitDataIds[0]
+            return visitDataId
+        else:
+            self.log.warning(f"Failed to find visitId for {dataId}, got {visitDataIds}. Do you need to run"
+                             " define-visits?")
+            return None
+
+    def safeWrite(self, object, datasetType, dataId):
```
mfisherlevine, line 1029: I think this function needs to be renamed: given that it actually can remove existing data this doesn't seem like the best words for it. `overwriteIfExisting` or something like that?

python/lsst/rubintv/production/rubinTv.py
yalsayyad: DM-37426 filed.

python/lsst/rubintv/production/rubinTv.py
yalsayyad: Agreed, but it's not going to get merged tonight if I do that now.

python/lsst/rubintv/production/rubinTv.py
yalsayyad: how about just `clobber`

PR #34, opened 2022-12-21
DM-37218: Bring cleanrooms online

mfisherlevine:


python/lsst/rubintv/production/slac/mosaicing.py
```diff
@@ -0,0 +1,352 @@
...
+    result : `lsst.pipe.base.Struct`
+        A pipeBase struct containing the ``output_mosaic`` as an
+        `lsst.afw.image.Image`.
+
+    Notes
+    -----
+    Tricks used for speed:
+        Pulling the detector names from a butler.get(component='detector')
+        takes ~8s for 189 CCDs. Using the dRef.dataId['detector'] and getting
+        the name from the camera is ~instant.
```
TallJimbo, line 191: That's the difference between doing I/O and not doing I/O!

python/lsst/rubintv/production/slac/utils.py
```diff
@@ -0,0 +1,162 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import time
+from time import sleep
```
TallJimbo, line 23: I'm a bit amused that you didn't just write `time.sleep` later.

python/lsst/rubintv/production/slac/utils.py
```diff
@@ -0,0 +1,162 @@
...
+    -------
+    detectorsWithData : `list` [`int`]
+        The list of detector numbers for which the dataset exists.
+    """
+    return sorted([det.id for det in butler.registry.queryDimensionRecords('detector',
+                                                                           dataId=expRecord.dataId,
+                                                                           datasets=dataset)])
+
+
+def waitForDataProduct(butler, expRecord, dataset, detector, timeout, cadence=1, logger=None):
```
TallJimbo, line 120: Could this be unified with other similar logic elsewhere, maybe in the `ButlerWatcher`?

python/lsst/rubintv/production/starTracker.py
```diff
@@ -141,15 +141,27 @@ class StarTrackerWatcher():
...
 
     def _getLatestImageDataIdAndExpId(self):
         """Get the dataId and expId for the most recent image in the repo.
+
+        Returns
+        -------
+        latestFile : `str`
+            The filename of the most recent file or ``None`` is nothing is
+            found.
+        dayObs : ``
```
TallJimbo, line 159:
```suggestion
        dayObs : `int` or `None`
```

python/lsst/rubintv/production/starTracker.py
```diff
@@ -141,15 +141,27 @@ class StarTrackerWatcher():
...
+            The dayObs of the ``latestFile`` or ``None`` is nothing is found.
+        seqNum : ``
```
TallJimbo, line 161:
```suggestion
        seqNum : `int` or `None`
```

python/lsst/rubintv/production/starTracker.py
```diff
@@ -141,15 +141,27 @@ class StarTrackerWatcher():
...
+            The seqNum of the ``latestFile`` or ``None`` is nothing is found.
+        expId : ``
```
TallJimbo, line 163:
```suggestion
        expId : `int` or `None`
```

python/lsst/rubintv/production/starTracker.py
```diff
@@ -141,15 +141,27 @@ class StarTrackerWatcher():
...
+        self.uploader = Uploader(bucketName)
         self.log = _LOG.getChild("watcher")
         self.heartbeater = None
 
     def _getLatestImageDataIdAndExpId(self):
         """Get the dataId and expId for the most recent image in the repo.
+
+        Returns
+        -------
+        latestFile : `str`
```
TallJimbo, line 156:
```suggestion
        latestFile : `str` or `None`
```

python/lsst/rubintv/production/uploaders.py
```diff
@@ -0,0 +1,271 @@
...
+            The channel name.
+        flatlinePeriod : `float`
+            The period after which to consider the channel dead.
+
+        Returns
+        -------
+        success : `bool`
+            Did the upload succeed?
+        """
+        filename = "/".join([self.HEARTBEAT_PREFIX, channel]) + ".json"
```
TallJimbo, line 74: I'm guessing `os.path.join` is inappropriate because `self.HEARBEAT_PREFIX` is a non-file URI?

python/lsst/rubintv/production/utils.py
```diff
@@ -50,6 +56,261 @@
...
+    def __post_init__(self) -> None:
+        # Touch the _config after init to make sure the config file can be
+        # read.
+        # Any essential items can be touched here, but note they must all
+        # exist in all the different locations, otherwise it will fail in some
+        # locations and not others, so add things with caution.
+        self._config
+
+    def _checkDir(self, dirName: str, createIfMissing=True) -> None:
+        """
```
TallJimbo, line 169: Forgotten docstring.

python/lsst/rubintv/production/utils.py
```diff
@@ -50,6 +56,261 @@
...
+                self.log.info(f'Directory {dirName} does not exist, creating it.')
+                os.makedirs(dirName, exist_ok=True)  # exist_ok necessary due to potential startup race
+        if not os.path.isdir(dirName):
+            msg = f"Directory {dirName} does not exist"
+            if createIfMissing:
+                msg += " and could not be created."
+            raise RuntimeError(msg)
+
+    def _checkFile(self, filename):
+        """
```
TallJimbo, line 185: Forgotten docstring.

python/lsst/rubintv/production/watchers.py
```diff
@@ -0,0 +1,226 @@
...
+        self.doRaise = doRaise
+        self.log = _LOG.getChild("fileWatcher")
+        self.doHeartbeat = False
+        if heartbeatChannelName:
+            self.doHeartbeat = True
+            self.heartbeatChannelName = heartbeatChannelName  # currently unused
+            self.heartbeater = Heartbeater(heartbeatChannelName,
+                                           self.locationConfig.bucketName,
+                                           self.HEARTBEAT_UPLOAD_PERIOD,
+                                           self.HEARTBEAT_FLATLINE_PERIOD)
```
TallJimbo, line 73: Since you don't seem to use `self.heartbeatChannelName` after this at all, it might be better to replace this with:
```
if heartbeatChannelName:
    self.heartbeater = Heartbeater(...)
else:
    self.heartbeater = None
```
and then use `if self.heartbeater is not None` instead of `self.doHeatbeat`.

I do think it'd be good to at least set `self.heartbeater` (and `self.heartbeatChannel`) to `None` in an `else` branch here.  Not having attributes at all sometimes doesn't feel right to me, though I don't have a concrete explanation for why `None` is better than leaving them undefined.

python/lsst/rubintv/production/watchers.py
```diff
@@ -0,0 +1,226 @@
...
+        for product in self.dataProducts:
+            # NB if you list multiple products for datasets= then it will only
+            # give expRecords for which all those products exist, so these must
+            # be done as separate queries
+            records = self.butler.registry.queryDimensionRecords("exposure",
+                                                                 datasets=product,
+                                                                 where=where)
+            records.order_by('-exposure.id')  # the minus means descending ordering
+            records.limit(1)
+            expRecordDict[product] = list(records)[0]
```
TallJimbo, line 198: Do you need to worry here about the query not returning anything?

python/lsst/rubintv/production/watchers.py
```diff
@@ -0,0 +1,226 @@
...
+                    for product, expRecord in found.items():
+                        writeDataIdFile(self.locationConfig.dataIdScanPath, product, expRecord, log=self.log)
+                        lastWrittenIds[product] = expRecord.id
+                    # beat after the callback so as not to delay processing
+                    # and also so it is only called if things are working
+                    # TODO: re-enable?
+                    # self.heartbeater.beat()
+
+            except Exception as e:
+                raiseIf(self.doRaise, e, self.log)
```
TallJimbo, line 226: This is a big exception block, and it looks like there might be ways in which the `doRaise=False` branch could lead to this really hammering the butler hard if something goes wrong, which might lead to new errors, and so on.  Maybe add a `sleep` there, or even inside `raiseIf` to make sure that happens anywhere it might be needed?

And I think a common pattern is to increase that sleep duration as the failures continue - I certainly see GMail doing that to me, though this is well outside my area of expertise.

scripts/meta/writeBotScripts.py
```diff
@@ -0,0 +1,46 @@
+import os
+from lsst.utils import getPackageDir
+
+isrRunnerScript = """
```
TallJimbo, line 4: if you're going to the trouble of putting a license in your generated script, you should put it in the script generator.

scripts/meta/writeBotScripts.py
```diff
@@ -0,0 +1,46 @@
...
+if os.path.isdir(outputDir) is False:
+    print(f"Output directory {outputDir} does not exist, creating it")
+    os.makedirs(outputDir)
+
+for detNum in range(205):
+    scriptName = os.path.join(outputDir, f"runIsrRunner_{detNum:03}.py")
+    with open(scriptName, 'w') as f:
+        contents = isrRunnerScript.format(detNum, detNum)
+        f.write(contents)
+        print(f"Wrote runner script for {detNum} to {scriptName}")
```
TallJimbo, line 46: I think a subcommand approach would organize things much better than script-generation, if that's something that still might be viable to change. And `argparse` makes it pretty easy (I would not recommend `click` for this).

```
bin/run-rubintv butler-watcher
bin/run-rubintv isr-runner 018
```
etc.

TallJimbo (approved):
I've left lots of little comments, and a few not-on-this-ticket ones, but overall looks fine to me.

I've intentionally not commented on:
 - TODO comments
 - lack of docstrings (unless it looked like you'd intended to write one and forgot based on context)
 - the scripts that appear to be machine-generated.

I'm definitely glad you didn't try to include black and isort before the review, and I definitely think you should make mypy a separate ticket from that (and start that with just getting mypy set up and checking some small fraction of it - it's easy to tell it to ignore code with no annotations, and that's absolutely what you should do at first).

python/lsst/rubintv/production/__init__.py
```diff
@@ -27,3 +27,5 @@
 from .monitorPlotting import *
 from .catchupService import *
 from .channels import *
+from .watchers import *
+from .metadataServers import *
```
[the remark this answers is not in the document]
mfisherlevine: Does that need or even want doing? I left that out deliberately, as nobody should be instantiating those classes directly, and I expected any files which wanted stuff from them to be doing relative imports to get them. Is that bad or unwise?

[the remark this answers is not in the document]
TallJimbo: I think it might confuse Sphinx, if you cared about running that, and I think it might be worth a code comment, but I don't think it's bad if it's intentional.

python/lsst/rubintv/production/metadataServers.py
```diff
@@ -0,0 +1,173 @@
...
+    locationConfig : `lsst.rubintv.production.utils.LocationConfig`
+        The location configuration.
+    metadataDirectory : `str`
+        The name of the directory for which the metadata is being served. Note
+    that this directory and the ``shardsDirectory`` are passed in because
+    although the ``LocationConfig`` holds all the location based path info (and
+    the name of the bucket to upload to), many directories containg shards
+    exist, and each one goes to a different page on the web app, so this class
+    must be told which set of files to be collating and uploading to which
+    channel.
```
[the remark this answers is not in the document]
mfisherlevine: Oh god, you're right - I thought this was the style for some reason, I wonder how many other places I've done that 😳

python/lsst/rubintv/production/slac/botTesting.py
```diff
@@ -0,0 +1,270 @@
...
+        Parameters
+        ----------
+        expRecord : `lsst.daf.butler.DimensionRecord`
+            The exposure record.
+        """
+        # run isr, calculate any numbers you want here with the raw
+        # put those into a json or pickle
+        # maybe create a thumbnail, or actually just the binned raw image
+        # and the scaling of it, so that it can be assembled into a focal
+        # plane thumbnail
```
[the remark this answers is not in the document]
mfisherlevine: Haha, that was actually my notes to myself when I was first imagining all this code from scratch 🙂 All gone now!

python/lsst/rubintv/production/slac/utils.py
mfisherlevine: Me too, was probably a VSCode auto-import or something. I've switched it back 🙂

python/lsst/rubintv/production/__init__.py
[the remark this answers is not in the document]
mfisherlevine: Oh, it was a passing thought, I'm happy to just add everything if that's the norm 👍

python/lsst/rubintv/production/watchers.py
mfisherlevine: I wouldn't have thought so, but as it turns out, yes, yes I do 🙂

mfisherlevine: (this is because of the way TS8 does its exposure ids, meaning they're a factor of 10 larger than on the summit) Also I need to protect for dataProducts which don't exist, so double-yes!

python/lsst/rubintv/production/highLevelTools.py
```diff
@@ -0,0 +1,407 @@
...
+    bucket = client.get_bucket(bucketName)
+    prefix = f'{channel}/{PREFIXES[channel]}'
+    blobs = list(bucket.list_blobs(prefix=prefix))
+
+    logger.info(f'Found {len(blobs)} for channel {channel} in bucket')
+
+    # names are like
+    # 'auxtel_monitor/auxtel-monitor_dayObs_2021-07-06_seqNum_100.png'
+    days = set([b.name.split(f'{prefix}_dayObs_')[1].split('_seqNum')[0] for b in blobs])
+    days = [int(d.replace('-', '')) for d in days]  # days are like 2022-01-02
```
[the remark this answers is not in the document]
mfisherlevine: Yeah, I basically need them spoon-feeding to me, so as you say, I think the main thing is to have it so that I can maintain it. Thanks though 🙂

python/lsst/rubintv/production/slac/botTesting.py
```diff
@@ -0,0 +1,270 @@
...
+        return 1
+    elif instrument == "LSSTCam":
+        return 201
+    elif instrument == "LSST-TS8":
+        return 9
+    else:
+        raise ValueError(f"Unknown instrument {instrument}")
+
+
+class RawProcesser():
```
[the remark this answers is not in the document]
mfisherlevine: It's quite possible it should, but I've not decided to do that yet while this bit is being so heavily developed, as these classes have just been born. It's quite possible I'll add that on a future ticket though 👍

python/lsst/rubintv/production/slac/botTesting.py
```diff
@@ -0,0 +1,270 @@
...
+            self.log.info(f'Wrote binned image for detector {detNum}')
+
+    def run(self):
+        """Run continuously, calling the callback method with the latest
+        expRecord.
+        """
+        self.watcher.run(self.callback)
+
+
+class Plotter():
```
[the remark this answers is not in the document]
mfisherlevine: For sure, but again as above.

python/lsst/rubintv/production/slac/utils.py
mfisherlevine: Uggg, you're right that that is _extremely_ similar to the `_waitForDataProduct` on the `BaseChannel`, this was part of the refactoring work. I think I will add a TODO to the code for now, and address that at the same time as making the new cleanroom code inherit from those base channels, as that all is part of the same coherent bit of work.

python/lsst/rubintv/production/uploaders.py
mfisherlevine: Correct, these aren't paths, these are things to use the the GCS bucket.

python/lsst/rubintv/production/watchers.py
mfisherlevine: Yeah, totally agree, not really sure why I did it like that tbh.

python/lsst/rubintv/production/watchers.py
mfisherlevine: Ah yes, good shout. And you're quite right, that pattern is, I believe, referred to as "exponential backoff" and is in fact implemented automatically in how the uploads are reattempted in failure in GCS (and in the `Uploader` class in this package). I don't think it's worth the hassle of coding it up here, but I've certainly added a sleep and code comment as to why it's there.

scripts/meta/writeBotScripts.py
mfisherlevine: lol, indeed

scripts/meta/writeBotScripts.py
mfisherlevine: This is complicated, and there are two reasons why I didn't do it. But firstly, yes, you're absolutely right, and we might still go that way. But that means a change in how the kubernetes pods which run these are made, and that's not something I have control over, and try to bug the people who do as little as possible. And secondly, given that I'm still not sure that each won't end up being somewhat hand-crafted (though to what extent I don't know yet, maybe config options per runner, maybe pods doing multiple detectors, but also maybe nothingetc) I thought it best just to keep everything in my control for now.

Also note that it's only the mass-produced pods which are like this, all the one-off ones are still handmade, so it would only be the `isr-runner detNum` ones I think, and I very much plan to keep it that way (lest I need 500 pods!)

PR #36, opened 2023-02-04
DM-37309: Bring NightReport channel online

mfisherlevine:
Also contains a few urgent fix-ups to StarTracker code which needed to be done on this ticket to same from a git nightmare on the summit.


python/lsst/rubintv/production/exposureLogUtils.py
```diff
@@ -0,0 +1,90 @@
...
+    parts = obsId.split('_')
+    if len(parts) != 4:
+        logger.warning(f'Failed to parse {obsId=}')
+        return None
+    seqNum = int(parts[3])
+    return seqNum
+
+
+def getLogsForDayObs(dayObs, logger=None):
+    """Get a dictionary of log messagse for the dayObs, keyed by seqNum.
```
TallJimbo, line 58:
```suggestion
    """Get a dictionary of log messages for the dayObs, keyed by seqNum.
```

python/lsst/rubintv/production/plotting/nightReportPlotBase.py
```diff
@@ -0,0 +1,282 @@
...
+        self.plotName = plotName
+        self.plotGroup = plotGroup
+        self.channelName = channelName
+        self.locationConfig = locationConfig
+        self.uploader = uploader
+        self.log = logging.getLogger(f'lsst.rubintv.production.nightReportPlots.{plotName}')
+
+    def getSaveFilename(self):
+        """XXX docs
+        """
```
TallJimbo, line 70: Looks like it should be addressed on this ticket.

python/lsst/rubintv/production/plotting/nightReportPlotBase.py
```diff
@@ -0,0 +1,282 @@
...
+            The front page metadata, as a dataframe.
+        ccdVisitTable : `pandas.DataFrame`
+            The visit summary table for the current day.
+
+        Returns
+        -------
+        success : `bool`
+            Did the plotting succeed, and thus upload should be performed?
+        """
+        raise NotImplementedError()
```
TallJimbo, line 161: I don't spot any difference between this method and its base class version, and if that's the case, you should just drop it.  Same for other intermediate base classes here.

python/lsst/rubintv/production/plotting/nightReportPlotBase.py
```diff
@@ -0,0 +1,282 @@
...
+
+        Parameters
+        ----------
+        nightReport : `lsst.rubintv.production.nightReport.NightReport`
+            The night report for the current night.
+        metadata : `pandas.DataFrame`
+            The front page metadata, as a dataframe.
+        ccdVisitTable : `pandas.DataFrame`
+            The visit summary table for the current day.
+        """
```
TallJimbo, line 177: No need to repeat docs from the base class if they're the same; they'll get inherited if there is no docstring.  Lots of cases where this applies through this PR.

python/lsst/rubintv/production/plotting/nightReportPlotBase.py
```diff
@@ -0,0 +1,282 @@
...
+        if not success:
+            self.log.warning(f'Plot {self.plotName} failed to create')
+            return
+
+        saveFile = self.getSaveFilename()
+        plt.savefig(saveFile)
+        plt.close()
+
+        # XXX need to add a renaming of the plot on upload as it now
+        # starts with the channelName which is no longer correct for upload        line too long to ensure I don't forget about this!
```
TallJimbo, line 191: Don't forget about this!

python/lsst/rubintv/production/plotting/latissNightReportPlots.py
```diff
@@ -0,0 +1,1000 @@
...
+        ax = plt.gca()
+        xfmt = md.DateFormatter('%m-%d %H:%M:%S')
+        ax.xaxis.set_major_formatter(xfmt)
+        ax.tick_params(which='both', direction='in')
+        ax.minorticks_on()
+        plt.legend()
+        return True
+
+
+class PsfE2Plot(LatissPlot):
```
TallJimbo, line 338: I'm not commenting about plotting code duplication for the most part on this PR, because I think fixing that kind duplication is often a deceptively hard problem, and very often the solutions are much worse than the problem.

But needing separate classes for plotting e1 and e2 is a pretty clear indication that something is wrong.  Happy to pair-code if you feel boxed in on this; it's not clear enough to me what led to this to propose a solution.

python/lsst/rubintv/production/plotting/latissNightReportPlots.py
```diff
@@ -0,0 +1,1000 @@
...
+        success : `bool`
+            Did the plotting succeed, and thus upload should be performed?
+        """
+        for item in ['Zeropoint', 'Filter']:
+            if item not in metadata.columns:
+                msg = f'Cannot create {self._PlotName} plot as required item {item} is not in the table.'
+                self.log.warning(msg)
+                return False
+
+        # TODO: get a figure you can reuse to avoid matplotlib memory leak
```
TallJimbo, line 90: You've got a lot of these comments, so I'll only comment on it once here:

- Instead of going all the way to a reusable `Figure`, it'd be worth trying to: get a `Figure` via `matplotlib.figure.Figure`, get an `Axes` (or a few) from `Figure.subplots`, and then call methods on `Axes` to do the plotting.  That requires far less global state (maybe none?) so it should be better at cleaning up after itself, and the `Axes` methods are almost all drop-in replacements for the `matplotlib.pyplot` functions (in m experience, you just have to put `set_` in front of `xlim`, `ylim`, `title`, and things like that).

- If that doesn't help, try `gc.collect().`

python/lsst/rubintv/production/plotting/latissNightReportPlots.py
```diff
@@ -0,0 +1,1000 @@
...
+
+        datesDict = nightReport.getDatesForSeqNums()
+
+        inds = metadata.index[metadata['Sky mean'] > 0].tolist()  # get the non-nan values
+        rawDates = np.asarray([datesDict[seqNum] for seqNum in inds])
+        bands = np.asarray(metadata['Filter'][inds])
+        # TODO: generalise this to all bands and add checks for if empty
+        rband = np.where(bands == 'SDSSr_65mm')
+        gband = np.where(bands == 'SDSSg_65mm')
+        iband = np.where(bands == 'SDSSi_65mm')
```
TallJimbo, line 167: It'd be good to add a little indirection here that lets you say `r`, `i`, or `z` here by looking up the full expected string in some kind of mapping, even if that's just a constant at the top of the file.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -795,18 +925,251 @@ def clobber(self, object, datasetType, dataId):
...
+
+class NightReportChannel(BaseButlerChannel):
+    """Class for running the AuxTel Night Report channel on RubinTV.
+
+    Parameters
+    ----------
+    locationConfig : `lsst.rubintv.production.utils.LocationConfig`
+        The locationConfig containing the path configs.
+    dayObs : `int`, optional
+        TODO: XXX document this!!! XXX
```
TallJimbo, line 976: Looks like it should be addressed.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -795,18 +925,251 @@ def clobber(self, object, datasetType, dataId):
...
+        if os.path.isfile(saveFile):
+            self.log.info(f'Resuming from {saveFile}')
+            self.report = NightReport(self.butler, self.dayObs, saveFile)
+            self.report.rebuild()
+        else:  # otherwise start a new report from scratch
+            self.report = NightReport(self.butler, self.dayObs)
+
+    def finalizeDay(self):
+        """XXX Docs
+        """
```
TallJimbo, line 1015: Looks like it should be addressed.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -795,18 +925,251 @@ def clobber(self, object, datasetType, dataId):
...
+        except Exception as e:
+            self.log.warning(f"Failed to load metadata table from {sidecarFilename}: {e}")
+            return None
+
+        if mdTable.empty:
+            return None
+
+        return mdTable
+
+    def getCcdVisitTable(self, dayObs):
```
TallJimbo, line 1054: Method name might a verb that implies more work being done than `get`.

python/lsst/rubintv/production/starTracker.py
```diff
@@ -479,3 +511,128 @@ def callback(self, filename):
...
+
+        All plots defined in __all__ in nightReportPlots are discovered,
+        created and uploaded. If any fail, the exception is logged and the next
+        plot is created and uploaded.
+        """
+        md = self.getMetadataTableContents()
+        self.log.info(f'Creating plots for dayObs {self.dayObs} with: '
+                      f'{0 if md is None else len(md)} items in the metadata table')
+
+        for plotClassName in starTrackerNightReportPlots.__all__:
```
TallJimbo, line 590: Having a dedicated module-scope variable for the plots to include here is probably better than `__all__`, even if you set one from the other in the module.  Even better would be a function in that module that could be called to create all of the plots - and that might provide the solution to the e1 vs e2 duplication I commented on elsewhere.

Same argument applies for LATISS plots, of course.

python/lsst/rubintv/production/starTracker.py
```diff
@@ -479,3 +511,128 @@ def callback(self, filename):
...
+                                 locationConfig=self.locationConfig,
+                                 uploader=self.uploader)
+                plot.createAndUpload(md)
+            except Exception:
+                self.log.exception(f"Failed to create plot {plotClassName}")
+                continue
+
+    def finalizeDay(self):
+        """XXX Docs
+        """
```
TallJimbo, line 604: Looks like it should be addressed.

TallJimbo (approved):
Looks good.  Once again I can't say I saw the forest, so I just reviewed trees, and while my comments are mostly minor many of them are in the, "I'm commenting here once for a ton of occurrences".

Biggest issue is that I think you need to change how `Plot` instances are constructed so they can take some custom parameters at construction.  I don't care if you have a fair amount of duplication in the plots because they _happen_ to have similar boilerplate, but I am bothered if the structure of the code forces you to duplicate even when you'd rather not, and I think that's what your construction pattern is doing.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -810,3 +814,99 @@ def clobber(self, object, datasetType, dataId):
...
+        self.dataProduct = 'quickLookExp'
+        self.uploader = Uploader()
+        self.butler = makeDefaultLatissButler()
+        self.log = _LOG.getChild("NightReportChannel")
+        self.channel = 'auxtel_night_report'
+        self.watcher = Watcher(self.dataProduct, self.channel)
+        self.doRaise = doRaise
+        self.dayObs = getCurrentDayObs_int()
+
+        # always attempt to resume on init
```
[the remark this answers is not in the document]
mfisherlevine: As discussed OOB, it's much easier to delete the file than to restart the process with different args (it's in k8s).

python/lsst/rubintv/production/plotting/nightReportPlotBase.py
mfisherlevine: Thanks - deleted.

python/lsst/rubintv/production/plotting/latissNightReportPlots.py
mfisherlevine: As discussed OOB, this can now be sorted quite easily, and will be addressed as needed on another ticket.

python/lsst/rubintv/production/plotting/latissNightReportPlots.py
mfisherlevine: This definitely needs doing, but I'll file a new ticket for it.

python/lsst/rubintv/production/plotting/nightReportPlotBase.py
mfisherlevine: All deleted, thanks.

python/lsst/rubintv/production/plotting/latissNightReportPlots.py
mfisherlevine: https://jira.lsstcorp.org/browse/DM-38287 and I've added a TODO to the code with that.

PR #37, opened 2023-03-09
DM-38303: Add catchup service for StarTrackers

mfisherlevine:


python/lsst/rubintv/production/starTracker.py
```diff
@@ -642,3 +676,165 @@ def callback(self, filename, doCheckDay=True):
...
+                    self.runCatchup()
+                    self.heartbeater.beat()
+                    lastRun = time.time()
+                    if hasDayRolledOver(self.dayObs):
+                        sleep(self.endOfDayDelay)  # give time for anything running elsewhere to finish
+                        self.runEndOfDay()  # sets new dayObs in a finally block
+                else:
+                    remaining = self.catchupPeriod - timeSince
+                    self.log.info(f'Waiting for catchup period to elapse, {remaining:.2f}s to go...')
+                    sleep(self.loopSleep)
```
czwa, line 835: Should this sleep for the larger of `remaining` and `loopSleep`?  Or is this potentially shorter to ensure that the heartbeat is sent on line 837?

scripts/summit/misc/runMisc.py
```diff
@@ -1,5 +1,8 @@
-import os
-dirName = os.path.dirname(__file__)
-fileToRun = os.path.join(dirName, '..', 'misc', 'runStarTrackerNightReport.py')
-print(f'Running {fileToRun}...')
-os.system(f'python {fileToRun}')
+# import os
```
czwa, line 1: Can you comment here on this and `runMisc2.py`?  I simply do not understand why these are here.

python/lsst/rubintv/production/starTracker.py
```diff
@@ -151,6 +153,30 @@ def dayObsSeqNumFromFilename(filename):
...
+        The dayObs.
+    seqNum : `int`
+        The seqNum.
+
+    Returns
+    -------
+    filename : `str`
+        The filename.
+    """
+    rootPath = '/project/GenericCamera/'
```
czwa, line 173: Below you have a `locationConfig`.  Should that be used here instead of hard coding the path?

czwa (approved):
A couple of typos, and a few "I don't understand" questions that I don't feel are blockers.

python/lsst/rubintv/production/starTracker.py
```diff
@@ -642,3 +676,165 @@ def callback(self, filename, doCheckDay=True):
...
+
+        seqNums = list(mdTable.index)
+        successfulFitColumn = 'Calculated Ra' + camera.suffixWithSpace
+        missing = [s for s in seqNums if mdTable[successfulFitColumn][s] is np.nan]
+        return missing
+
+    def catchupCamera(self, camera):
+        """Catch up a single camera.
+
+        TODO: add a way of recording fails and skipping them in future so
```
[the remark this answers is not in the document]
mfisherlevine: DM-38313

scripts/summit/misc/runMisc.py
mfisherlevine: Of course! So, every time I add a new running service a new pod has to be created to run it in, each has an entry point that is one of these scripts. This requires manual intervention in the kubernetes recipe and redeployment by Tiago, who's not always available immediately. Having a couple of miscellaneous pods sitting around, paused, which I can redirect to run any new service I might cook up, is incredibly useful. So the workflow is that I can use these to run one of the other scripts which hasn't been added to the k8s recipe yet, and then when he gets around to it, adds a new entry point, and these go back to being spare.

mfisherlevine: I don't know if that was terribly clear, but if not I'm happy to explain with words on Zoom or something.

mfisherlevine: And while we don't usually leave stuff commented out, given that it's doing nothing, and makes redirection much easier for next time they come into use, I tend to just leave the cruft there, ready to be updated and reactivated.

python/lsst/rubintv/production/starTracker.py
mfisherlevine: Admittedly that's not ideal, but actually the `locationConfig` only holds  the `starTrackerDataPath: '/project'` part, the `GenericCamera` part is hardcoded elsewhere, and soon all this will be butler driven once Robert gets that stuff all working, so this part in particular is very temporary, so I don't think it's worth the hassle of getting the location config in there right now.

python/lsst/rubintv/production/starTracker.py
mfisherlevine: Yeah, to be honest all that sleeping stuff is pretty rough, and just makes sure we don't unnecessarily hammer things. The point isn't to count all the way down in one go though, for the exact reason you said.

scripts/summit/misc/runMisc.py
czwa: Can you add a comment to that effect to the files?  It'll save any future reviewer from needing to wonder about them.

PR #38, opened 2023-03-22
DM-38312: Fix minor bug in starTracker catchup when cols don't exist

mfisherlevine:


python/lsst/rubintv/production/starTracker.py
```diff
@@ -737,6 +737,11 @@ def getMissingImageSeqNums(self, camera):
 
         seqNums = list(mdTable.index)
         successfulFitColumn = 'Calculated Ra' + camera.suffixWithSpace
+        if successfulFitColumn not in mdTable.columns:
```
arunkannawadi, line 740: Is it possible to do this instead?

```python
try:
    missing = [s for s in seqNums if mdTable[successfulFitColumn][s] is np.nan]
except KeyError:
    self.log.debug("%s does not exist ", successfulFitColumn)
    return [ ]
```

mfisherlevine: I think that's _possible_, yes, but usually using exceptions for control flow is discouraged. Is there an advantage to doing it this way that you're thinking of?

arunkannawadi: Question is whether this is really a control flow or not. If you expect that `successfulFitColumn` wouldn't exist in `mdTable` fairly regularly, or if the absence of the column is expected for some reason, then `if`-`else` block is fine.

mfisherlevine: It is expected, and the comment about thrashing above is saying exactly that. If they take a test image early in the day which won't solve (no stars!) then the table exists but there are no solves, and that stays like that until they go on sky.

PR #40, opened 2023-04-04
DM-38446: Fix mixed use of butler dis and yaml files

mfisherlevine:


config/config_slac_testing.yaml
```diff
@@ -36,8 +36,8 @@ astrometryNetRefCatPath: '/sdf/home/m/mfl/u/astrometry_net'
 #######################
 
 # Butler paths - use full paths not aliases here as path existence is checked
-ts8ButlerPath: '/sdf/group/rubin/repo/ir2'
-botButlerPath: '/sdf/group/rubin/repo/main'
+ts8ButlerPath: '/sdf/group/rubin/repo/ir2/butler.yaml'
+botButlerPath: '/sdf/group/rubin/repo/main/butler.yaml'
```
hsinfang, line 40: Maybe use this chance to make everything consistent by also updating `botButlerPath` in `config_slac.yaml` ?

hsinfang: Looking at `config_slac.yaml`'s `botButlerPath`, I guess it wasn't used because I don't think it'd work. It probably should be `/repo/main/butler.yaml`

hsinfang (approved):

config/config_slac_testing.yaml
mfisherlevine: Reply to first comment: yes, good shout, doing so.
Reply to second comment: As discussed OOB, this does need fixing, but not to that, it should be the same: `/repo/ir2/butler.yaml`
Reply to third comment: Yes, doing that too, thanks!

timj: The funny thing is I tried to design things such that people didn't need to use the trailing "butler.yaml" at all...

PR #41, opened 2023-04-05
DM-38353: Fix nans in RubinTV table

mfisherlevine:


python/lsst/rubintv/production/utils.py
```diff
@@ -490,6 +492,32 @@ def catchPrintOutput(functionToCall, *args, **kwargs):
...
+def sanitizeNans(obj):
+    """Recursively sanitize an object of any NaN-valued items.
+
+    Nans are not JSON de-serializable, so this function replaces them with
+    ``None``.
+
+    Parameters
+    ----------
+    obj : `object`
+        Any object to sanitize.
```
hsinfang, line 504: Maybe add a comment clarifying that this expects a `list` or `dict` or  individual object, but not any iterable?

hsinfang (approved):

PR #44, opened 2023-04-28
DM-38947: Use seq_num directly from exposurelog

hsinfang:


mfisherlevine (approved):
Look wonderful, and massive bonus points for unit tests using mock messages! 👏

PR #45, opened 2023-05-01
DM-38958: Fix bug in shard merging

mfisherlevine:


python/lsst/rubintv/production/metadataServers.py
```diff
@@ -126,7 +126,9 @@ def mergeShardsAndUpload(self):
             if shard:  # each is a dict of dicts, keyed by seqNum
                 for seqNum, seqNumData in shard.items():
                     seqNumData = sanitizeNans(seqNumData)  # remove NaNs
-                    data.update({seqNum: seqNumData})
+                    if seqNum not in data.keys():
+                        data[seqNum] = {}
```
isullivan, line 130: I'd suggest adding a warning of some kind that `seqNum` was not in `data`.

isullivan (approved):
Please add a warning or log message to notify that `seqNum` was missing from `data`

python/lsst/rubintv/production/metadataServers.py
mfisherlevine: That's actually a totally normal part of the process. When a shard for a new sequence number is inserted into the main `data` dict, there won't be an entry yet, so it needs an empty dict making, in order to be able to call `update()` on it, so it really doesn't warrant either a warn or even an info level log - it's happening constantly. Does that make sense?

PR #52, opened 2023-05-18
DM-39266: Fix gains for ComCam and LSSTCam

mfisherlevine:


tests/test_slac_utils.py
```diff
@@ -0,0 +1,78 @@
...
+        self.assertEqual(len(ts8Gains), 9)
+        for ccd, gains in ts8Gains.items():
+            self.assertEqual(len(gains), 16)
+
+        # check we have all the gains for ComCam
+        comCamGains = getGains('LSSTComCam')
+        self.assertEqual(len(comCamGains), 9)
+        for ccd, gains in comCamGains.items():
+            self.assertEqual(len(gains), 16)
+
```
hsinfang, line 44: I find it somewhat weird that the unit tests are verifying the data, but otherwise okay.

hsinfang (approved):

tests/test_slac_utils.py
mfisherlevine: In principle, I totally agree! The reason I put this in though, was just that up until now I was applying the TS8 gains to all instruments (albeit basically deliberately, as it was only coded up for TS,8 and then I used the same code for the other instruments) so I just added this to confirm that I wasn't still accidentally serving the same sets of gains to all instruments.

PR #54, opened 2023-05-25
DM-39407: Improve some log outputs

mfisherlevine:


python/lsst/rubintv/production/metadataServers.py
```diff
@@ -129,13 +131,18 @@ def mergeShardsAndUpload(self):
...
+                    updating.add((dayObs, seqNum))
             os.remove(shardFile)
 
             with open(mainFile, 'w') as f:
                 json.dump(data, f)
             if not isFileWorldWritable(mainFile):
                 os.chmod(mainFile, 0o777)  # file may be amended by another process
 
+        if updating:
+            for (dayObs, seqNum) in sorted(updating, key=lambda x: (x[0], x[1])):
```
hsinfang, line 143: Our [dev guide](https://developer.lsst.io/python/style.html#lambda-should-not-be-used) advises against lambda

hsinfang (approved):

python/lsst/rubintv/production/metadataServers.py
mfisherlevine: As discussed OOB, I believe the intent there is to avoid assigning to lambda, which I very very much agree with for many reasons. I think this case is one of the few places where we would actually recommend it though.

PR #57, opened 2023-06-27
DM-39698: Add time-based replotting

mfisherlevine:


python/lsst/rubintv/production/slac/botTesting.py
```diff
@@ -393,19 +397,22 @@ def writeRebHeaderShard(self, expRecord, raw):
 
         if self.instrument == 'LSST-TS8':
             detectorList = TS8_REB_HEADER_DETECTORS
+            rebNumber = detector.getName().split('_')[1][1]  # This is the X part of the S part of R12_SXY
```
taranu, line 400: Not that it really matters, but just FYI you can do `.split('_', 1)`.

python/lsst/rubintv/production/slac/botTesting.py
```diff
@@ -393,19 +397,22 @@ def writeRebHeaderShard(self, expRecord, raw):
...
+            itemName = f'REB{rebNumber} Header'
         elif self.instrument == 'LSSTComCam':
```
taranu, line 402: Why not merge this into one `if self.instrument == 'LSST-TS8' or self.instrument == 'LSSTCam':` statement, if they're identical?

python/lsst/rubintv/production/slac/botTesting.py
```diff
@@ -661,7 +682,7 @@ def getInstrumentChannelName(instrument):
             case _:
                 raise ValueError(f'Unknown instrument {instrument}')
 
-    def callback(self, expRecord, doPlotMosaic=False, doPlotNoises=False):
+    def callback(self, expRecord, doPlotMosaic=False, doPlotNoises=False, timeout=5):
```
taranu, line 685: Missing docstring for timeout.

python/lsst/rubintv/production/slac/botTesting.py
```diff
@@ -704,75 +725,89 @@ def run(self):
...
+        records = {r: files for r, files in records.items() if r.timespan.end < mostRecentExp.timespan.end}
         return records
 
     def run(self):
         """Run continuously, looking for complete file sets and plotting them.
         """
-        # TODO: turn this into an end-of-day service, checking for day rollover
-        # and calling findLeftoverMosaics() with findIncomplete=True, and set
-        # deleteRegardless when plotting them to tidy up at the end of the day.
+        # workloads is a list of tuples, each of which is: (a filter function,
```
taranu, line 769: This is where I say this should definitely be a (data)class.

python/lsst/rubintv/production/utils.py
```diff
@@ -791,7 +791,9 @@ def getShardedData(path,
                                            seqNum=seqNum)
 
     start = time.time()
-    while time.time() - start < timeout:
+    firstLoop = True
+    while firstLoop or time.time() - start < timeout:
```
taranu, line 795: I generally prefer having lengthier conditions in parentheses like `(time.time() - start < timeout)` for clarity, even if it's not necessary. YMMV.

taranu (approved):
Seems fine to me. Take the comments/suggestions as you will.

python/lsst/rubintv/production/slac/botTesting.py
mfisherlevine: They're actually not quite identical - different detector lists for the different instruments. Going to merge this as-is here, but just FYI that all this stuff will get ripped out and done properly in DM-39922

python/lsst/rubintv/production/slac/botTesting.py
mfisherlevine: Yup, definitely agree. Could you just take a quick look at `c7882a5` and check that's what you meant? I think that making the items `Callable` and making the call to `partial()` in making the dataclasses is the right way to go, so that the calling code is basically unchanged, as opposed to putting the args and kwargs in the dataclass, but just wanted to check this is what you were thinking.

python/lsst/rubintv/production/utils.py
mfisherlevine: Totally agree 👍

PR #59, opened 2023-07-21
DM-40144: Add parallel upload of all-sky movies and stills for EPO

mfisherlevine:


python/lsst/rubintv/production/allSky.py
```diff
@@ -321,6 +321,7 @@ def __init__(self, *,
                  outputImageDir,
                  outputMovieDir,
                  uploader,
+                 epoUploader,
```
hsinfang, line 324: Docstring for this new parameter?

python/lsst/rubintv/production/allSky.py
```diff
@@ -412,7 +414,12 @@ def animateFilesAndUpload(self, isFinal=True):
                 raise RuntimeError(f'Failed to find movie {creationFilename}')
 
         if not self.DRY_RUN:
-            self.uploader.googleUpload(self.channel, creationFilename, uploadAsFilename)
+            self.uploader.googleUpload(self.channel, creationFilename, uploadAsFilename, isLargeFile=True)
```
hsinfang, line 417: Is adding the `isLargeFile=True` part an independent change from adding epoUploader?

python/lsst/rubintv/production/allSky.py
```diff
@@ -436,6 +443,12 @@ def uploadLastStill(self, convertedFiles):
             self.uploader.googleUpload(channel=channel,
                                        sourceFilename=sourceFilename,
                                        uploadAsFilename=uploadAsFilename)
+            try:
```
hsinfang, line 449: This part looks like an independent change from adding epoUploader too?

hsinfang (approved):

python/lsst/rubintv/production/allSky.py
mfisherlevine: Yup, good shout, thanks.

python/lsst/rubintv/production/allSky.py
mfisherlevine: It's just a bugfix really. I realised I should be using it for the EPO uploader when I wrote the line, and spotted that I didn't have it set for the normal upload, so just set it there too. It'll go away soon enough though, as all the normal "upload" stuff will be S3 instead of GCS soon.

python/lsst/rubintv/production/allSky.py
mfisherlevine: I think this comment is outdated now, but GitHub somehow isn't showing that to be the case. I think you're referring to the fact that this used to be using the main uploader to send the still, which was wrong. It's now using the EPO uploader to send the still to them as well, and so is part of the change, I just don't know why GitHub hasn't caught that that's been updated already (maybe because the comment is on the `try` line and that line itself didn't change, I guess). But yes, it's all part of the same thing.

PR #64, opened 2023-08-16
DM-40411: Fix minor bug in replotter

mfisherlevine:


python/lsst/rubintv/production/slac/botTesting.py
```diff
@@ -560,7 +560,7 @@ def __init__(self, butler, locationConfig, instrument, doRaise=False):
                                    doRaise=doRaise)
         self.fig = plt.figure(figsize=(12, 12))
         self.doRaise = doRaise
-        self.STALE_AGE = 5*60  # in seconds, so 5 mins
+        self.STALE_AGE_SECONDS = 45  # in seconds
```
hsinfang, line 563: Would it be useful to make this a config and not hard-coded?

hsinfang (approved):

python/lsst/rubintv/production/slac/botTesting.py
mfisherlevine: I don't think so - I'm just tuning it at the moment, but there will probably be a "right" answer quite soon (5 mins was right before, but I think 45s should be a pretty good setting with these changes applied). Also, the problem with configs for k8s deployed code is that it's just as hard to set the parameter as it is to change the hard-coded value on some level, if not even harder with my helm/k8s/argo/kustomize etc skills!

PR #67, opened 2023-08-18
DM-40439: Add timestamps to all sky images

mfisherlevine:


python/lsst/rubintv/production/allSky.py
```diff
@@ -104,7 +108,70 @@ def dayObsFromDirName(fullDirName, logger):
...
+        if dateTimeStr:
+            # dateTimeStr comes out like "2021:08:17 06:57:00"
+            dateStr, timeStr = dateTimeStr.split(' ')
+            year, month, day = dateStr.split(':')
+            dateStr = f"{year}-{month}-{day}"
+            return dateStr, timeStr
+        else:
+            if not logger:  # only create if needed
+                logger = _LOG.getChild("getDateTimeFromExif")
+                logger.warning(f"Failed to get DateTime from exif data in {filename}")
```
czwa, line 170: I think you want this line outside of the `if` block.

python/lsst/rubintv/production/allSky.py
```diff
@@ -376,13 +476,14 @@ def convertFiles(self, files, forceRegen=False):
         for file in sorted(files):  # sort just helps debug
             outputFilename = self._getConvertedFilename(file)
             self.log.debug(f"Converting {file} to {outputFilename}")
+            date, time = getDateTimeFromExif(file, logger=self.log)
```
czwa, line 479: Won't this will raise a `TypeError` if the `DateTime` entry is missing from the EXIF data? It will return `None`, which my quick test tells me "is not iterable."

python/lsst/rubintv/production/allSky.py
```diff
@@ -104,7 +108,70 @@ def dayObsFromDirName(fullDirName, logger):
...
+        The path to the Ubuntu bold font, or ``None`` if not found.
+    """
+    fontPaths = matplotlib.font_manager.findSystemFonts(fontpaths=None, fontext='ttf')
+
+    ubuntuBoldPath = [f for f in fontPaths if f.find('Ubuntu-B.') != -1]
+    if not ubuntuBoldPath:
+        if not logger:  # only create if needed
+            logger = _LOG.getChild("getUbuntuFontPath")
+        logger.warning('Warning - cound not fund Ubuntu bold font!')
+        return
```
czwa, line 131: In this case, will the annotation code simply use a default system font?

czwa (approved):
A couple of minor points that might cause edge-case failures.

python/lsst/rubintv/production/allSky.py
mfisherlevine: For better and for worse - yes. It's different on every system I've tried it on, and sometimes it's awful, but this way failover is just a bad font, not failure, so it seemed like the best way to go.

python/lsst/rubintv/production/allSky.py
mfisherlevine: Good point, thanks!

python/lsst/rubintv/production/allSky.py
mfisherlevine: I don't ever expect them to be missing, but this is a good point. I've changed it to return two empty strings so that the interface is always the same, and to only pass them through for adding if they contain text, so that should sort that out.

PR #74, opened 2024-01-23
DM-41232: Add ICS hardpoint plots to TMA channel

mfisherlevine:


config/config_slac_testing.yaml
```diff
@@ -66,3 +66,8 @@ comCamBinning: 8
 # # paths for serving TS8 metadata at SLAC
 comCamMetadataPath: '/sdf/home/m/mfl/u/rubintv/LSSTComCam/sidecar_metadata'
 comCamMetadataShardPath: '/sdf/home/m/mfl/u/rubintv/LSSTComCam/sidecar_metadata/shards'
+
+
+# summit-like configs
+tmaMetadataPath: '/sdf/home/m/mfl/u/rubintv/tma/sidecar_metadata'
```
czwa, line 72: Since these are similar to those for ComCam, I assume there's a ticket to migrate these to a user-agnostic location already.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -1259,14 +1261,155 @@ def __init__(self, *,
...
+            'measuredForceMean1': m1m3IcsResult.stats.measuredForceMean1,
+            'measuredForceMean2': m1m3IcsResult.stats.measuredForceMean2,
+            'measuredForceMean3': m1m3IcsResult.stats.measuredForceMean3,
+            'measuredForceMean4': m1m3IcsResult.stats.measuredForceMean4,
+            'measuredForceMean5': m1m3IcsResult.stats.measuredForceMean5,
+        }
+
+        # do the max of the absolute values of the forces
+        md['M1M3 ICS Hardpoint AbsMax-Max Force'] = max(m1m3ICSHPMaxForces.values(), key=abs)
+        md['M1M3 ICS Hardpoint AbsMax-Mean Force'] = max(m1m3ICSHPMeanForces.values(), key=abs)
```
czwa, line 1372: I did not know about this trick.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -1277,85 +1420,28 @@ def processDay(self, dayObs, skipEventNumbers={}):
...
-
-                filename = self._getSaveFilename(dayObs, event)
-                self.figure.savefig(filename)
-                self.uploader.uploadPerSeqNumPlot(self.plotChannelName,
-                                                  dayObsInt=dayObs,
-                                                  seqNumInt=event.seqNum,
-                                                  filename=filename,
-                                                  isLiveFile=False
-                                                  )
+                rowData = self.runMountMotionAnalysis(event)
```
czwa, line 1423: I suspect this is the "unscalable plots" you were talking about.  These could be merged into a helper function, and probably should be if the number of plots expands much beyond 2.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -1386,8 +1473,9 @@ def run(self):
 
                 # TODO: need to work out a better way of dealing with pod
                 # restarts. At present this will just remake everything.
-                justMade = self.processDay(dayObs, skipEventNumbers=plotted)
-                plotted.update(justMade)
+                mountMotionJustPlotted, icsJustPlotted = self.processDay(dayObs, skipEventNumbers=plotted)
+                mountMotionPlotted.update(mountMotionJustPlotted)
```
czwa, line 1477: Do these sets get used anywhere?  Something certainly should replace `plotted` on line 1476, but I'm not sure if it should be one of these, or the union of the two.

czwa (approved):
I think other than the question of L1476 of rubinTV.py, this looks fine to me.

config/config_slac_testing.yaml
mfisherlevine: Not a ticket, at such, but yes, it's a bit piece of tech-debt in RA. I'll file it now though, for sorting out LocationConfig class entirely.

mfisherlevine: DM-42912

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: Neat, eh? I learned this recently too 🙂

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: Yeah, exactly. The whole thing needs a refactor: DM-42911

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: Oh, that's a great catch, thank you. And uggg, you're right, technically it's neither - this is why it needs refactoring. Will take a look into that now.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -1402,46 +1411,55 @@ def runM1M3HardpointAnalysis(self, event):
...
-
-            try:
-                rowData = self.runM1M3HardpointAnalysis(event)
-            except Exception as e:
-                rowData = {event.seqNum: {'ICS processing error?': '😔'}}
-                self.log.exception(f'Failed to plot event {event.seqNum}')
-                raiseIf(self.doRaise, e, self.log)
-            finally:
-                icsPlotted.add(event.seqNum)  # don't retry plotting on failure
+            if event.seqNum not in self.plotsMade['MountMotionAnalysis']:
```
czwa, line 1439: Following from the above question, it seems like this will retry any events that already had no results, and will attempt to plot MountMotionAnalysis plots for events that did have M1M3HardpointAnalysis results and vice versa.  I think the fails are fast, and the day_obs reset will ensure things are flushed, but I just wanted to check that this is intentional.

czwa (approved):
Nothing breaking, but a few questions so I'm understanding correctly.

python/lsst/rubintv/production/rubinTv.py
```diff
@@ -1402,46 +1411,55 @@ def runM1M3HardpointAnalysis(self, event):
...
-                continue
             assert event.dayObs == dayObs
+
+            nMountMotionPlots = len(self.plotsMade['MountMotionAnalysis'])
+            nM1M3HardpointPlots = len(self.plotsMade['M1M3HardpointAnalysis'])
+            # the interesting phrasing in the message is because these plots
+            # don't necessarily exist, due either to failures or M1M3 analyses
+            # only being valid for some events so this is to make it clear
+            # they've been processed.
+            self.log.info(f'Found {len(events)} events for {dayObs=} of which '
```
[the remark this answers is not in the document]
mfisherlevine: So `getEvents` returns all events for the day, and actually does the calculation/generation, so that `len(events)` will be all the events for the current day, and so is strictly ≥ either of those other numbers.

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: As discussed OOB, all is well here.

PR #79, opened 2024-02-26
DM-43032: Stop rewriting row-shard for processed events

mfisherlevine:


python/lsst/rubintv/production/rubinTv.py
```diff
@@ -1427,6 +1427,14 @@ def processDay(self, dayObs):
         """
         """
         events = self.eventMaker.getEvents(dayObs)
+
+        # check if every event seqNum is in both the M1M3HardpointAnalysis and
+        # MountMotionAnalysis sets, and if not, return immediately
+        if (all([event.seqNum in self.plotsMade['MountMotionAnalysis'] for event in events]) and
+                all([event.seqNum in self.plotsMade['M1M3HardpointAnalysis'] for event in events])):
+            self.log.info(f'No new events found for {dayObs} (currently {len(events)} events).')
```
isullivan, line 1435: You might consider raising the severity to `warning`, if that wouldn't cause too much noise.

isullivan (approved):
Looks good

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: Ha, actually, if it moves in any direction I think that would go to `DEBUG` level - this is something that comes every few seconds while the TMA is stationary, which is sometimes for days on end. I thought about it and decided to leave it as INFO though, just so that you can be certain that the service is still running and checking for new events.

mfisherlevine: Moreover though, it's a totally normal and expected part of it operating.

PR #82, opened 2024-03-05
DM-43169: Check for timebase errors on AuxTel images

mfisherlevine:


python/lsst/rubintv/production/rubinTv.py
```diff
@@ -58,6 +58,7 @@
                                         getAzimuthElevationDataForEvent,
                                         )
 from lsst.summit.utils.efdUtils import clipDataToEvent, makeEfdClient
+from lsst.summit.utils.auxtel.mount import hasTimebaseErrors
```
isullivan, line 61: Minor point, but add this import alphabetically.

isullivan (approved):

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: I mean, the rest are a total mess, but sure 😄 (`isort` and `black` coming in the fairly near term for this stuff too btw)

PR #85, opened 2024-03-07
DM-43213: Fix broken logic introduced by DM-43141

mfisherlevine:


python/lsst/rubintv/production/starTracker.py
```diff
@@ -873,6 +878,7 @@ def catchupCamera(self, camera, dayObs):
                                                 cameraType=camera.cameraType)
         for file in filenames:
             self.log.info(f'Catching up {file}')
+            gc.collect()
```
isullivan, line 881: Is this still necessary, after the matplotlib `plot` changes?

python/lsst/rubintv/production/starTracker.py
```diff
@@ -399,6 +401,7 @@ def __init__(self,
         self.solver = CommandLineSolver(indexFilePath=self.locationConfig.astrometryNetRefCatPath,
                                         checkInParallel=True,
                                         timeout=30)
+        self.fig = plt.figure(figsize=(16, 16))
```
isullivan, line 404: Is there ever a case where you would want to pass an existing figure reference into this class?

python/lsst/rubintv/production/starTracker.py
```diff
@@ -50,6 +51,7 @@
 from .uploaders import Uploader, Heartbeater
 from .baseChannels import BaseChannel
 from .plotting import starTrackerNightReportPlots
+import gc
```
isullivan, line 54: If you do still need this, move it up to the first block of imports.

isullivan (approved):
Looks good!

python/lsst/rubintv/production/starTracker.py
mfisherlevine: You know, probably not, but I don't think it does any harm. Then again, I do dislike code which scatters those around for no reason, so maybe I will removed and see if it's all OK with it actually, I can always put it back, but if I leave it it'll likely live on forever.

python/lsst/rubintv/production/starTracker.py
mfisherlevine: No that I can think of tbh.

PR #87, opened 2024-03-12
DM-43269: Refactor StarTracker code

mfisherlevine:


python/lsst/rubintv/production/starTracker.py
```diff
@@ -446,16 +446,22 @@ def runAnalysis(self, exp, filename):
         if not np.isfinite(pressure):
             self.log.warning("Pressure not found in header, falling back nominal value=0.770 bar")
             pressure = 0.770
+        else:
+            pressure = None  # if it's valid, do not override it
```
laurenam, line 454: I'm confused by the logic here.  As is, it seems that if `pressure` is `np.finite`, you override it to be `None` (for the others below as well).

python/lsst/rubintv/production/starTracker.py
```diff
@@ -62,7 +62,11 @@
 __all__ = (
     'StarTrackerWatcher',
     'StarTrackerChannel',
-    'StarTrackerNightReportChannel'
+    'StarTrackerNightReportChannel',
+    'StarTrackerCatchup',
+    'getCurrentRawDataDir',
+    'getDataDir',
+    'getFilename',
```
laurenam, line 69: Bubble up the last three for appearance-in-file ordering?

laurenam (approved):

python/lsst/rubintv/production/starTracker.py
mfisherlevine: Ah, I deliberately put them in the order they're defined in the code to make it easier to keep them correct. I wonder if that's not right now though. Thoughts?

python/lsst/rubintv/production/starTracker.py
mfisherlevine: The `None` is the default value for the function it's passed to, _i.e._ telling it to take it from the `visitInfo`. If you look inside that function, the `overrideX` params allow you to manually specify something, and that will always be preferentially taken if it's supplied (but it's unitless, so it's dangerous, which is why, if it's already present, I allow that function to get it itself, rather than supplying it as an override). So yes, `None` is supplied to the override params.

I guess another way of doing it would be to make an kwarg dict, and add these items to it when it's not present/finite, and then just unpack that into the `getAltAzFromSkyPosition` function, which would mean that I don't have to pass `None` in explicitly. It's more obvious/natural in some ways, but making a dict and then `**`ing it is less obvious in others. If that made sense, any thoughts/preferences on the approach?

mfisherlevine: I think I've probably persuaded myself to do it the kwarg dict way actually, 1 sec.

python/lsst/rubintv/production/starTracker.py
laurenam: I'm I misreading something?  The first function I see below is `getCurrentRawDataDir`.  Anyway, I do leave it up to you how you'd like this ordered (the only guidance I see in the dev guide is https://developer.lsst.io/python/numpydoc.html#complete-example-module

python/lsst/rubintv/production/starTracker.py
mfisherlevine: OK, see here:
https://github.com/lsst-sitcom/rubintv_production/pull/87/commits/c783cbde65ff1fbf7de704322a5c329f48884502

python/lsst/rubintv/production/starTracker.py
mfisherlevine: lol, you're not misreading anything, I'm just an idiot 🙃 Fixed.

PR #96, opened 2024-04-30
DM-44127: Add S3 uploader connection checker

kkeijuro:
Create utility function which just checks that the thing is properly connected, with the correct credentials and profiles available, etc, because as it is at the moment, it's possible to make one S3 Uploader, and only realise that it doesn't actually work when you try to write something

### DIFF
--- a/python/lsst/rubintv/production/uploaders.py
+++ b/python/lsst/rubintv/production/uploaders.py
@@ -268,11 +268,22 @@ class MultiUploader(IUploader):
     def __init__(self):
         # TODO: thread the remote upload
         self.localUploader = createLocalS3UploaderForSite()
+        localOk = self.localUploader.checkAccess()
+        if not localOk:
+            raise RuntimeError("Failed to connect to local S3 bucket")
 
         try:
             self.remoteUploader = createRemoteS3UploaderForSite()
         except Exception:
             self.remoteUploader = None
+
+        remoteRequired = getSite() in ['summit', 'tucson', 'base']
+        if remoteRequired and not self.hasRemote:
+            raise RuntimeError("Failed to create remote S3 uploader")
+        elif remoteRequired and self.hasRemote:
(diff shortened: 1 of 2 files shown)


python/lsst/rubintv/production/uploaders.py
```diff
@@ -493,6 +495,62 @@ def uploadPerSeqNumPlot(
...
+        with tempfile.NamedTemporaryFile() as testFile:
+            testFile.write(testContent)
+            testFile.flush()
+            try:
+                self._s3Bucket.upload_file(testFile.name, fileName)
+            except (BotoCoreError, ClientError) as e:
+                self._log.exception(f"S3Uploader Write Access check failed: {e}")
+                return False
+
+        _, fixedFilePath = tempfile.mkstemp()
```
mfisherlevine, line 530: If you change this to using a `TemporaryFile` or `NamedTemporaryFile` [context manager](https://docs.python.org/3/library/tempfile.html#module-tempfile) you can simplify it quite a bit, including removing the whole `finally` block.

python/lsst/rubintv/production/uploaders.py
```diff
@@ -493,6 +495,62 @@ def uploadPerSeqNumPlot(
...
+        """
+        testContent = b"Connection Test"
+        dateStr = datetime.now().strftime('%Y%m%d_%H%M%S')
+        fileName = f'test/{tempFilePrefix}_{dateStr}_file.txt'
+        with tempfile.NamedTemporaryFile() as testFile:
+            testFile.write(testContent)
+            testFile.flush()
+            try:
+                self._s3Bucket.upload_file(testFile.name, fileName)
+            except (BotoCoreError, ClientError) as e:
```
mfisherlevine, line 526: You need to add a catch for `S3UploadFailedError` here too.

mfisherlevine (commented):
Fixed a few things, added something, will merge now.

PR #101, opened 2024-10-09
DM-45573: Adding magnitude limit (and updating effTime) column mappings

kadrlica:
Needs to wait until the rest of DM-45573 goes into production.


kadrlica:
@mfisherlevine I think this is ready for review (or maybe you've gone ahead and done something different instead). It'd be awesome if we can backfill these quantities.

rhiannonlynne:
Agree that these would be useful! 
Currently the consdb has "None" values for these columns.

mfisherlevine:
Merged and live on the summit - sorry for the extreme delay here 😔

PR #104, opened 2024-11-26
DM-47801: Add time since previous exposure to RubinTV table

mfisherlevine:


python/lsst/rubintv/production/oneOffProcessing.py
```diff
@@ -166,14 +166,34 @@ def calcPsfAndWrite(self, exp: Exposure, dayObs: int, seqNum: int) -> None:
             self.log.info(f"Wrote measured PSF for {dayObs}-{seqNum} det={exp.detector.getId()}: {fwhm:.3f}")
         except Exception as e:
             self.log.error(f"Failed to calculate PSF for {dayObs}-{seqNum} det={exp.detector.getId()}: {e}")
-            if self.doRaise:
-                raise
```
isullivan, line 170: Did you mean to remove these lines? If so, it would be helpful to put them in a separate commit.

python/lsst/rubintv/production/processingControl.py
```diff
@@ -668,6 +668,10 @@ def dispatchGatherSteps(self, who: str) -> bool:
             if who == "SFM":
                 (expRecord,) = self.butler.registry.queryDimensionRecords("exposure", dataId=dataCoords[0])
                 self.dispatchOneOffProcessing(expRecord, PodFlavor.ONE_OFF_CALEXP_WORKER)
+            if who == "AOS":
+                for dataCoord in dataCoords:
+                    (expRecord,) = self.butler.registry.queryDimensionRecords("exposure", dataId=dataCoord)
+                    self.dispatchOneOffProcessing(expRecord, PodFlavor.ONE_OFF_POSTISR_WORKER)
```
isullivan, line 674: This is doing much more than just calculating the exposure timing, isn't it?

isullivan (approved):
Please double-check that you meant to remove the `doRaise` block in the first commit, and that the second commit doesn't have unintended side effects.

python/lsst/rubintv/production/processingControl.py
mfisherlevine: It is, yeah. That should have been being done to those exposures all along really, and I hadn't thought about out how to do it, but realised it was actually pretty easy 🙂 You were just checking your understanding and that I was doing what I intended, right? Or were you thinking the PR should maybe be renamed because of this?

python/lsst/rubintv/production/oneOffProcessing.py
mfisherlevine: Yeah, the whole `doRaise` pattern isn't really being adhered to in this new module tbh, so I think I'll take this as the poke I needed to make it adhere to that, thanks.

PR #108, opened 2024-12-02
DM-47607: Add a one-off exposure record processor

mfisherlevine:


python/lsst/rubintv/production/utils.py
```diff
@@ -1349,3 +1349,29 @@ def removeDetector(dataCoord: DataCoordinate, butler: Butler) -> DataCoordinate:
...
+    """
+    filterMap = {
+        "u_02": "u_color",
+        "g_01": "g_color",
+        "r_03": "r_color",
+        "i_06": "i_color",
+        "z_03": "z_color",
+        "y_04": "y_color",
+    }
+    return filterMap.get(physicalFilter)
```
TallJimbo, line 1377: Is https://github.com/lsst/utils/blob/e8ea38501a61ba974ac8dd10b8c611a05c43cd42/python/lsst/utils/plotting/figures.py#L65-L102 relevant?

TallJimbo (approved):

python/lsst/rubintv/production/utils.py
mfisherlevine: Very much so, though it doesn't want to be directly linked. The colour flags on RubinTV are indeed derived from those values, but the frontend doesn't expect hex codes, it's semantic, so that we can change things on the frontend as needed without changing the backend code.

PR #112, opened 2025-02-20
DM-49051: Update pipelines for CalibrateImageTask

mfisherlevine:


mfisherlevine:
Stuck a bunch of linter fixes on this because the new env necessitated them since you reviewed, I trust that's OK.

PR #113, opened 2025-02-20
DM-48411: Use squid proxy instead of LHN annotation for USDF uploads

mfisherlevine:


leeskelvin (approved):
Also LGTM! Thanks for confirming the new summit proxy default is expected.

PR #126, opened 2025-06-19
DM-51465: Write zernikes to ConsDB

mfisherlevine:


python/lsst/rubintv/production/pipelineRunning.py
```diff
@@ -607,6 +626,59 @@ def postProcessVisitSummary(self, quantum: Quantum) -> None:
...
+        visitRecord = dRef.dataId.records["visit"]
+        assert visitRecord is not None, "visitRecord is None, this shouldn't be possible"
+        data = getEfdData(self.efdClient, "lsst.sal.MTRotator.rotation", expRecord=visitRecord)
+        physicalRotation = np.nanmean(data["actualPosition"])
+
+        detector = dRef.dataId.records["detector"]  # not a detector object, but a detector dimension
+        assert detector is not None, "detector is None, this shouldn't be possible"
+        detectorId: int = detector.id  # hence .id rather than .getId()
+
+        zkTable = zkTable[zkTable["label"] == "average"]
```
isullivan, line 665: Please add a comment addressing why you are downselecting `zkTable` in this way.

python/lsst/rubintv/production/pipelineRunning.py
```diff
@@ -607,6 +626,59 @@ def postProcessVisitSummary(self, quantum: Quantum) -> None:
...
+        zkColsHere = [col for col in zkTable.colnames if col.startswith("Z")]
+        nollIndicesHere = np.asarray([int(col.removeprefix("Z")) for col in zkColsHere], dtype=int)
+        # Grab Zernike values, convert to dense array, save
+        zkSparse = zkTable[zkColsHere].to_pandas().values[0]
+        zkDense = makeDense(zkSparse, nollIndicesHere, MAX_NOLL_INDEX)
+        rotationMatrix = zernikeRotMatrix(MAX_NOLL_INDEX, -np.deg2rad(physicalRotation))
+        zernikeValues: np.ndarray = zkDense / 1e3 @ rotationMatrix[4:, 4:]
```
isullivan, line 672: Why are the lower values dropped? No need to address if this should be obvious to someone working with this code, but it's not apparent to a newbie.

isullivan (approved):
Code looks good, and well documented. My only complaint is about the protected import of `makeDense` from `ts_wep`. Could that one function be moved to a different package that is a common dependency of both RA and T&S instead?

python/lsst/rubintv/production/pipelineRunning.py
```diff
@@ -607,6 +626,59 @@ def postProcessVisitSummary(self, quantum: Quantum) -> None:
         except Exception:
             self.log.exception("Failed to populate visit row in ConsDB")
 
+    def postProcessCalcZernikes(self, quantum: Quantum) -> None:
+        """Post-process the Zernike table to send results to ConsDB."""
+        # protect import to stop the whole package depending on ts_wep. If this
+        # becomes a problem we could copy the functions or just accept that RA
+        # needs T&S software.
+        from lsst.ts.wep.utils.zernikeUtils import makeDense
```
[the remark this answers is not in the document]
mfisherlevine: I agree it's not great, and/but there's precedent for that below, which both means that we're not making anything _worse_, but conversely this is the 2nd time we've needed T&S code outside. But refactoring that is a major pain, and so something I'm currently still punting on... (having the build stacks be separate is honestly one of the biggest "original sins" of the project, but until it gets a little worse than this I don't think it's time to solve them).

python/lsst/rubintv/production/pipelineRunning.py
gmegh: We are selecting the average zernike coefficients among all the donuts of that specific detector. `zkTable` provides the estimate for each donut pair, but it suffices to take the average among donut pairs of that detector, for the metrics we want to keep in consDB. If someone wants more details they can query the butler.

python/lsst/rubintv/production/pipelineRunning.py
gmegh: Because we are only really keeping track of z4 and above. You could add a comment there about this.

mfisherlevine: Thanks - will add a comment. ConsDB only has slots for z4 upwards anyway, so that's justification enough 🙂

PR #130, opened 2025-07-03
DM-51714: Fix rendering bug from using class variables for figures

mfisherlevine:


craiglagegit:
Looks good to me.  Just a simple fix.

PR #178, opened 2026-04-13
DM-54577: Refactor package and update redis data structures

mfisherlevine:


architecture/architecture.md
```diff
@@ -0,0 +1,444 @@
...
+and "ISR". All share the same per-exposure tracking hash. The head node
+calls `dispatchGatherSteps()` for each one every loop iteration.
+
+### Handling Failures and Edge Cases
+
+- **Dispatch failures**: if no worker exists for a detector, that detector
+  is removed from the expected set via `removeExpectedDetectors()`. This
+  prevents the gather step from waiting forever.
+
+- **Worker failures**: if a worker crashes mid-processing, it still calls
```
fred3m, line 381: It seems to assume that a worker failure is always due to a processing error and not a possible hardware. Is that a safe assumption?

.claude/skills/rapid-analysis-architecture-sync/SKILL.md
```diff
@@ -0,0 +1,61 @@
+---
+name: rapid-analysis-architecture-sync
+description: Keep the architecture/ docs in sync with code changes in the rapid analysis backend. Use this skill whenever a change touches the distributed system's shape — Redis keys/formats/TTLs, pod flavors or queue naming, pipeline stages or task chaining, the gather/fanout mechanism, head-node or worker event loop logic, focal plane control or detector fanout, payload serialization, or CI test phases. The architecture/*.md files are the canonical source of truth for the system's design; letting them drift from the code is the single biggest way this project accumulates onboarding debt, so update the relevant doc in the same commit as the code change — not "later". Also apply when the user asks to "document this change", "update the architecture", or reviews a PR that obviously needs doc updates.
```
fred3m, line 3: I like this a lot. I think that I might implement something similar in scarlet lite.

.claude/skills/rapid-analysis-code-style/SKILL.md
```diff
@@ -0,0 +1,101 @@
+---
+name: rapid-analysis-code-style
+description: Apply the rapid analysis Python code style — camelCase identifiers, PascalCase classes, built-in type hints with "| None" (never typing.Optional/List/Dict), numpydoc docstrings with types, and black/isort formatting at line-length 110. Use this skill whenever writing or editing any Python code in this repo (adding a function, method, class, refactoring, renaming, writing a test). The conventions here diverge deliberately from PEP 8 (camelCase is the house style, inherited from the wider LSST stack) and from typing conventions common elsewhere in Python — do not "fix" them to PEP 8 even if they look unusual.
```
fred3m, line 3: We should probably make a skill (or even a tool) for the DM-style guide used for all projects.

.claude/skills/rapid-analysis-code-style/SKILL.md
```diff
@@ -89,12 +89,14 @@ def myFunction(param1: int, param2: str | None = None) -> bool:
...
-- Pre-commit hooks: trailing whitespace, YAML, isort, black, flake8
-- Build: LSST SCons + pyproject.toml
+## Tooling
+
+- **flake8** — linter; runs automatically via pre-commit.
+- **black** and **isort** — formatters; run automatically via pre-commit.
+- **mypy** — type checker; **not** run by pre-commit or CI, so you must run
+  it by hand on any Python change before declaring a task done. See the
+  `rapid-analysis-testing` skill for the command and the validation loop.
+- Build: LSST SCons + pyproject.toml.
```
fred3m, line 99: Great idea, I should do this too.

fred3m (approved):
I highly approve of using Claude to write the commits. After reading the design docs in full, I had Claude summarize all of the changes. Everything seems to make sense to me, and reading through the commit history was helpful in understanding what changed. My only curiosity is whether or not the code is functional at each commit. What I mean is did you make a bunch of changes and then have Claude cherry-pick commits or are these the ~77 different steps, each one leaving the code functional at that point? This is an important distinction, because if the code in most of the commits is not functional then this might be a bit of overkill.

I also like that you implemented the tests before all of the refactoring changes. That's the right way to do this and should give you confidence that you understand the effects of all of the changes made.

architecture/architecture.md
mfisherlevine: Not super safe, no (it ~never hardware, but it can be _e.g._ OOM kills). DM-51466 is already filed, and next on the list 🙂

.claude/skills/rapid-analysis-code-style/SKILL.md
mfisherlevine: For sure! This is the prototype 🙂 Once this is merged, and starts getting real usage, we can start thinking about how to centralise these!

mfisherlevine:
To answer the big question: no, these were pretty much all done as we went, so (with maybe a few exceptions, not sure, but _certainly_ fewer exceptions than if I'd written it all myself!) the code should be ~functional at each step. There's probably a few caveats, especially as the tests were being written, but it's more or less commit-as-we-iterated-and-developed.

PR #179, opened 2026-04-17
DM-54697: Process daytime checkout darks

mfisherlevine:


kfanning (approved):
Looks good to me. I prefer this implementation where that specific program can be a pass though, we could flexibly change this to a DOME image or similar.

mfisherlevine:
Agreed that hard-coding it isn't great. There's nowhere to pass it through from at present though. If we end up needing to keep adding to that list, I'll think about a redesign for sure.

PR #183, opened 2026-04-28
DM-54805: Docker build improvements + Fixes for USDF

ctslater:


python/lsst/rubintv/production/processingControl.py
```diff
@@ -732,7 +733,9 @@ def getLatestRunAndPrep(self, forceNewRun: bool) -> str:
             return newCollection
 
         allRunNums = [
-            int(run.removeprefix(self.outputChain + "/")) for run in allRuns if self.outputChain in run
+            int(re.match(self.outputChain + r"/(\d{1,3})$", run).group(1))
+            for run in allRuns
+            if re.match(self.outputChain + r"/(\d{1,3})$", run)
```
mfisherlevine:
```
pattern = re.compile(rf"{re.escape(self.outputChain)}/(\d+)$")
allRunNums = [int(m.group(1)) for run in allRuns if (m := pattern.match(run))]
```
Suggest that instead so that we escape the prefix and also don't limit ourselves to 3 digit runs?

mfisherlevine (commented):
Not going to claim I grok the eups stuff, but the rest looks fine, and I'll trust you that that part is right 🙂

PR #191, opened 2026-05-13
DM-54933: Fix all sky camera processing and add tests

mfisherlevine:


tests/test_allSky.py
```diff
@@ -0,0 +1,111 @@
...
+from unittest.mock import MagicMock
+
+import lsst.utils.tests
+from lsst.rubintv.production.allSky import DayAnimator
+from lsst.rubintv.production.uploaders import MultiUploader, Uploader
+
+TEST_DATA_DIR = os.path.join(os.path.dirname(__file__), "data", "all_sky")
+
+
+class AllSkyPipelineTestCase(lsst.utils.tests.TestCase):
```
ctslater: I don't think we need to use any lsst-specific test harnesses.

ctslater (approved):

tests/test_allSky.py
mfisherlevine: You know, I've always just copied and pasted that, and never really stopped to wonder what that actually gives us. What's the point of doing that, and of removing it?

ctslater: Minimizing dependencies on non-builtin-packages is good.

PR #193, opened 2026-05-14
DM-54818: Extend test coverage

mfisherlevine:


python/lsst/rubintv/production/rubinTv.py
```diff
@@ -84,366 +69,6 @@ def _catchPrintOutput(functionToCall: Callable, *args: Any, **kwargs: Any) -> st
     return f.getvalue()
 
 
-class CalibrateCcdRunner(BaseButlerChannel):
```
fred3m, line 82: Normally this would be deprecated. Is the argument here that no one downstream is using rubintv so it's ok to just delete?

fred3m (approved):
LGTM. The only thing that I flagged is that you removed some classes and functions that normally would be deprecated properly. If rubintv is special enough that it doesn't need deprecation that's fine, I just wanted to make sure this was a conscious decision.

python/lsst/rubintv/production/rubinTv.py
mfisherlevine: Yes, exactly - RA is an app, not a library, so doesn't do deprecation warnings etc, as there is no downstream.

PR #197, opened 2026-05-27
DM-55079: Improve butler polling for speed

mfisherlevine:


erykoff (approved):
I think this is fine, but is there any problem in doing the "full hammer" (unbounded) when we are down for multiple days?  If so I guess having an adaptive cadence might help, but that might be a premature optimization.

PR #209, opened 2026-06-25
DM-55094: Add per-image versioning to RA

mfisherlevine:


python/lsst/rubintv/production/consdbUtils.py
```diff
@@ -822,3 +842,64 @@ def populateMountErrors(
...
+    def populatePackageVersions(
+        self,
+        expRecord: DimensionRecord,
+        packageVersions: PackageVersions,
+        allowUpdate: bool = False,
+    ) -> bool:
+        """Write the tracked package versions for an exposure to ConsDB.
+
+        Writing lives here, in Rapid Analysis, rather than in summit_utils, so
+        that ordinary summit_utils users only ever see a way to *read* the
```
ctslater: What should prevent users from writing to consdb is not having credentials; hiding functions in different packages is very roundabout.

python/lsst/rubintv/production/consdbUtils.py
```diff
@@ -822,3 +842,64 @@ def populateMountErrors(
...
+        versions, never to write them. The blob shape (``toDict``) and the
+        table/column constants still come from summit_utils, and the read there
+        uses the same ones, so the write and the read cannot drift.
+        Used by the backfill tooling (see
+        ``highLevelTools.backfillPackageVersions``), whose source is the merged
+        AOS metadata written at processing time. Location-gated like the other
+        populate methods; the write is a direct, synchronous insert, since the
```
ctslater: Lot of claude adding superfluous smart-sounding adjectives here.

python/lsst/rubintv/production/consdbUtils.py
```diff
@@ -822,3 +842,64 @@ def populateMountErrors(
...
+        value is a JSON blob (not the scalar columns the shared insert helper
+        is typed for) and package versions are only ever written by backfill.
+
+        Parameters
+        ----------
+        expRecord : `lsst.daf.butler.DimensionRecord`
+            The exposure record to write the versions for. The dataId and the
```
ctslater: "The dataId ..." is all claude explanation to the author not docstrings.

python/lsst/rubintv/production/highLevelTools.py
```diff
@@ -766,6 +768,74 @@ def backfillVisit1QuicklookForDay(
...
+    sidecarFile = os.path.join(metadataPath, f"dayObs_{dayObs}.json")
+    with open(sidecarFile) as f:
+        metadata = json.load(f)
+
+    where = f"exposure.day_obs={dayObs} AND instrument='{instrument}'"
+    records = butler.query_dimension_records("exposure", where=where)
+    recordsBySeqNum = {record.seq_num: record for record in records}
+
+    nWritten = 0
+    for seqNumStr, cells in metadata.items():
```
ctslater: I would call this a record or a row instead of `cells`.

python/lsst/rubintv/production/highLevelTools.py
```diff
@@ -766,6 +768,74 @@ def backfillVisit1QuicklookForDay(
...
+        shardDict = cells.get(PACKAGE_VERSIONS_SHARD_KEY)
```
ctslater: what does this have to do with shards? Isn't this the versions dictionary?

python/lsst/rubintv/production/packageVersions.py
```diff
@@ -0,0 +1,552 @@
...
+        packageNames = TRACKED_PACKAGES
+    if installedPackageNames is None:
+        installedPackageNames = TRACKED_INSTALLED_PACKAGES
+    versions = {name: getPackageVersion(name) for name in packageNames}
+    for name in installedPackageNames:
+        versions[name] = getInstalledPackageVersion(name)
+    return PackageVersions(versions=versions)
+
+
+def makePackageVersionShardDict(packageVersions: PackageVersions) -> dict[str, str]:
```
ctslater: Consider this a PR-wide comment: "shard" doesn't seem to be relevant here and seems to be something more like a display-friendly version? I actually can't tell what it means here, but it doesn't have anything to do with sharding.

python/lsst/rubintv/production/packageVersions.py
```diff
@@ -0,0 +1,552 @@
...
+    """Render package versions as a metadata-shard cell.
+
+    The returned dict is the cell value written to the AOS metadata page. Apart
```
ctslater: A good docstring would say something like "The result is a dict with entries DISPLAY_VALUE of blah and a VERSIONS key that is blah". Comparing it to other things makes me reverse engineer what the output actually is.

python/lsst/rubintv/production/parsers.py
```diff
@@ -75,6 +75,13 @@ def sanitizeNans(obj: Any) -> Any:
     scientific notation) are converted to floats to ensure proper JSON numeric
     typing.
 
+    Book cells - dicts carrying a ``DISPLAY_VALUE`` marker, rendered behind a
```
ctslater: This whole comment is using a ton of words to say that entries containing the key DISPLAY_VALUE are ignored. 

But what does `DISPLAY_VALUE` mean? Is that a new concept that's being introduced in this PR?

python/lsst/rubintv/production/processingControl.py
```diff
@@ -737,6 +746,73 @@ def __init__(
...
+        # dispatch loop never pays for the git calls, and cross-check them
+        # against the Dockerfile (advisory only, never fatal).
+        self.log.info(
+            f"Tracked package versions (hash {self.packageVersions.versionHash()}):"
+            f" {self.packageVersions.versions}"
+        )
+        self._checkPackageVersionsAgainstDockerfile()
+
+    @functools.cached_property
+    def packageVersions(self) -> PackageVersions:
```
ctslater: Why call this instead of calling `getCurrentPackageVersions()`?

python/lsst/rubintv/production/packageVersions.py
```diff
@@ -0,0 +1,552 @@
...
+    """
+    if packageNames is None:
+        packageNames = TRACKED_PACKAGES
+    return [name for name in packageNames if not os.environ.get(envVarForPackage(name))]
+
+
+def findDockerfile() -> str | None:
+    """Locate the Dockerfile shipped alongside the rubintv_production checkout.
+
+    The Dockerfile is copied into the image at the root of the
```
ctslater: If the Dockerfile is leaking into the image that's a misfeature.

python/lsst/rubintv/production/packageVersions.py
```diff
@@ -0,0 +1,552 @@
...
+    pattern = re.compile(rf"^\s*{re.escape(packageName)}==?([^\s\\]+)")
+    for line in lines:
+        match = pattern.match(line)
+        if match:
+            return match.group(1)
+    return None
+
+
+def parseDockerfileRefs(dockerfilePath: str, condaPackages: list[str] | None = None) -> dict[str, str]:
+    """Parse the package version pins out of the Dockerfile.
```
ctslater: Parsing the dockerfile itself seems really suspect; you can do whatever you want at docker build time so why not write some file with the specific versioning information you want to store, in a better format than dockerfiles.

python/lsst/rubintv/production/packageVersions.py
```diff
@@ -0,0 +1,552 @@
...
+        pin = _findCondaPin(lines, name)
+        if pin is not None:
+            refs[name] = pin
+    return refs
+
+
+def versionsMatch(gitVersion: str, dockerfileRef: str) -> bool:
+    """Check whether a git version agrees with a Dockerfile ref.
+
+    A tag pin matches exactly. A SHA pin could be abbreviated in the
```
ctslater: "A tag pin matches exactly" feels like half a sentence.

python/lsst/rubintv/production/packageVersions.py
```diff
@@ -0,0 +1,552 @@
...
+    Dockerfile, so an abbreviated ref that prefixes the full git SHA counts as
+    a match.
+
+    Parameters
+    ----------
+    gitVersion : `str`
+        The version git reports (an exact tag or a full SHA).
```
ctslater: Somehow the phrase "exact tag" shows up a bunch here but it's just the same as a tag. There's no "inexact tag".

python/lsst/rubintv/production/packageVersions.py
```diff
@@ -0,0 +1,552 @@
...
+
+    Returns
+    -------
+    missing : `list` [`str`]
+        The package names whose directory env var is unset or empty, in the
+        order given.
+    """
+    if packageNames is None:
+        packageNames = TRACKED_PACKAGES
+    return [name for name in packageNames if not os.environ.get(envVarForPackage(name))]
```
ctslater: One line function that gets used in one place? I would inline this.

python/lsst/rubintv/production/packageVersions.py
```diff
@@ -0,0 +1,552 @@
...
+# than being git checkouts. These have no ``*_DIR`` env var, so their version
+# comes from the installed distribution metadata instead. In the Dockerfile
+# they are pinned with a conda ``name=version`` spec rather than an ``ARG
+# <name>_ref``.
+TRACKED_INSTALLED_PACKAGES = ["danish"]
+
+# The AOS metadata-shard cell the package versions are written under. Shared by
+# the writer (the head node) and the reader (the ConsDB backfill) so the two
+# can never drift apart.
+PACKAGE_VERSIONS_SHARD_KEY = "Package versions"
```
ctslater: Shard key?

ctslater (approved):

python/lsst/rubintv/production/consdbUtils.py
mfisherlevine: Yeah, totally agree creds are what actually stop people writing, and I shouldn't have made it sound like a safety thing. It's more that loads of our ConsDB writing already lives in RA (all the `ConsDBPopulator` stuff), and summit_utils is really the read-only side, so putting the write next to the other writes felt like the more natural pattern. Happy to chat if you feel strongly though!

python/lsst/rubintv/production/consdbUtils.py
mfisherlevine: Done, trimmed right down.

python/lsst/rubintv/production/consdbUtils.py
mfisherlevine: Gone.

python/lsst/rubintv/production/highLevelTools.py
mfisherlevine: Done, it's `row` now.

python/lsst/rubintv/production/highLevelTools.py
mfisherlevine: Yep, you're right. Renamed to `displayDict`, and the key's now `PACKAGE_VERSIONS_METADATA_KEY`.

python/lsst/rubintv/production/packageVersions.py
mfisherlevine: Renamed the constant and the dict helpers, thanks. The one I kept is `writePackageVersionShard`, because that really does write a metadata shard (via `writeMetadataShard`), which the TimedMetadataServer merges later, same as the other shards.

python/lsst/rubintv/production/packageVersions.py
mfisherlevine: Rewritten to say what's actually in the dict.

python/lsst/rubintv/production/parsers.py
mfisherlevine: Trimmed the comment right down. `DISPLAY_VALUE` isn't new, it's an existing RubinTV frontend convention: if a table cell's value is a dict, the frontend shows the `DISPLAY_VALUE` glyph (📖, ⏱ etc.) and clicking it pops out the rest of the dict. We already use it for the ISR configs, the perf timings, the M1M3 forces and so on. The change here just skips numeric coercion inside those dicts, so version strings like `"1.1"` don't get turned into floats.

python/lsst/rubintv/production/processingControl.py
mfisherlevine: Done, it's assigned directly in `__init__` now.

python/lsst/rubintv/production/packageVersions.py
mfisherlevine: It's not really leaking, it's just part of the repo checkout. The main place we look for it is the USDF CI, which runs straight from a git checkout, so it's there anyway. Even if we switched to a versions file we'd still need to find the pins somewhere, so I don't think this part changes.

python/lsst/rubintv/production/packageVersions.py
mfisherlevine: Fair. A build-time versions file might be nicer for some of this, but the main user of this check is the USDF CI. That runs from a plain checkout with no docker anywhere, so the Dockerfile is the only place the pins exist there. Checking what's pinned against what's actually running is the bit I really don't want to lose. It only ever warns, never raises, and there's a unit test that parses the real Dockerfile, so if the format drifts it'll fail in CI rather than going quiet.

python/lsst/rubintv/production/packageVersions.py
mfisherlevine: Reworded.

python/lsst/rubintv/production/packageVersions.py
mfisherlevine: Ha, fair. Removed.

python/lsst/rubintv/production/packageVersions.py
mfisherlevine: Done, inlined.

python/lsst/rubintv/production/packageVersions.py
mfisherlevine: Renamed to `PACKAGE_VERSIONS_METADATA_KEY`.

PR #211, opened 2026-06-30
DM-55340: Thread ConsDB writes so they don't block processing

mfisherlevine:


python/lsst/rubintv/production/consdbUtils.py
```diff
@@ -158,11 +159,126 @@ def changeType(key: str, typeMapping: dict[str, str]) -> Callable[[int | float],
...
+        # read whose result is needed to coerce values before the row can be
+        # built, so it cannot be made fire-and-forget like a write. But table
+        # schemas are static for the lifetime of a pod, so fetching once and
+        # caching keeps the read off the processing thread for every call
+        # after the first — the same goal the background writer serves for the
+        # writes. Only successful fetches are cached, so a transient failure is
+        # retried. Populated and read on the calling (processing) thread only.
+        self._schemaCache: dict[tuple[str, str], dict[str, str]] = {}
+
+    def flush(self, timeout: float | None = None) -> None:
```
bbrondel, line 221: These two functions look to be a halfway done implementation for a SIGTERM handler? I'd suggest creating a ticket to finish the SIGTERM handler. Alternatively, maybe just drop these functions as they are not used?

bbrondel (approved):
I'm really happy to see this improvement. Thank you for your efforts!

python/lsst/rubintv/production/consdbUtils.py
mfisherlevine: They are indeed the first step in that implementation. Right now, not a single pod in RA has a SIGTERM handler, and perhaps I should add one (it might make RA rollouts quicker), but I thought at least making clean shutdown _possible_ was better than not, especially as a new user within RA (e.g. one in `highLevelTools.py` perhaps) might choose to call these at a later date, so even though they're not used at the moment, I do see a place that might call them very soon, so I'd rather not drop them (unless they're wrong in some way).

mfisherlevine: I've filed DM-55356 to at least capture the vague intent.

